PR #17, opened 2019-01-31
Update README.md with cleaner alternative

benjie:
I fear the previous example may lead to bad code practices for those reading who opt not to adopt slonik's safer API. A more standard way of dealing with this is to *not* catch the error, and instead to use the `finally` block to guarantee the release of the connection. The main caveat here is that you must remember to `await` the promise, for that reason I didn't simplify it further to:

```js
try {
  return await connection.query(sql`SELECT produce_error()`);
} finally {
  await connection.release();
}
```

because I was worried people might drop the await from the return since that's generally safe (just not in try/catch/finally).

---

I understand you want to demonstrate that the old approach is bad, so perhaps invoking a transaction to cause nesting may make it clearer without compromising code quality - it's important that the `finally` block only contains one statement in case the statement itself raises an error and prevents the connection release:

```js
const connection = await pool.connect();
try {
  await connection.query(sql`BEGIN`);
  try {
    await connection.query(sql`SELECT do_something()`);
    const lastExecutionResult = await connection.query(sql`SELECT produce_error()`);
    return lastExecutionResult;
  } finally {
    await connection.query(sql`COMMIT`);
  }
} finally {
  await connection.release();
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,15 +190,10 @@ const main = async () => {
 
   try {
     lastExecutionResult = await connection.query(sql`SELECT produce_error()`);
-  } catch (error) {
+    return lastExecutionResult;
+  } finally {
     await connection.release();
-
-    throw error;
   }
-
-  await connection.release();
-
-  return lastExecutionResult;
 };
 
 ```


gajus:
Thank you for highlighting.

TIL: `finally` without `catch` does not swallow the error.

PR #37, opened 2019-02-22
fix: export createTypeParserPreset from root index

sebald:
This is part of the documentation:

```js
import {
  createTypeParserPreset
} from 'slonik';

createPool('postgres://', {
  typeParsers: [
    ...createTypeParserPreset()
  ]
});
```

But `createTypeParserPreset` is not part of the root `index.js`. This PR adds it to the exports.


gajus:
Thank you

PR #51, opened 2019-04-13
Date parameters

turuslan:
1. Inconsistent `Date` handling
Added test case showing that `sql` tag throws error if there is `Date` value, while `sql.valueList` doesn't.
```javascript
sql`${new Date()}` // throws
sql`${sql.valueList([new Date()])}` // returns { sql: '$1', type: Symbol(SLONIK_SQL), values: [ <date object> ] }
```

2. Treat `Date` as primitive value
Underlying "node-postgres" library [supports values of type `Date`](https://github.com/brianc/node-postgres/blob/master/lib/utils.js#L60).
Added `Date` to `PrimitiveValueExpressionType` flow type.
Added check to `isPrimitiveValueExpression` function.

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -299,7 +299,7 @@ export type AssignmentListTokenType = {|
   +type: typeof ComparisonPredicateTokenSymbol
 |};
 
-export type PrimitiveValueExpressionType = string | number | boolean | null;
+export type PrimitiveValueExpressionType = string | number | boolean | Date | null;
 
 export type SqlTokenType =
   ArraySqlTokenType |
--- a/src/utilities/isPrimitiveValueExpression.js
+++ b/src/utilities/isPrimitiveValueExpression.js
@@ -1,5 +1,5 @@
 // @flow
 
 export default (maybe: *): boolean %checks => {
-  return typeof maybe === 'string' || typeof maybe === 'number' || typeof maybe === 'boolean' || maybe === null;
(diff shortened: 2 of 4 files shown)


turuslan:
My changes are not related to [this linting error travis reports](https://travis-ci.org/gajus/slonik/jobs/519728197#L697).
Also, this error was reported when I tried to run lint original code before any changes.

turuslan:
I reproduced that linting error on unmodified code.
I forked repo and run travis.
The same hash b7b016b that was successfully built yesterday.
https://travis-ci.com/turuslan/slonik-travis/jobs/192737559#L684

turuslan:
Rebased on latest commit 82c7e23

gajus:
What is the value in supporting `Date` instance over just using numeric timestamp?

turuslan:
Writing `Date` conversion manually is error-prone.
Also, this conversion is already implemented and tested in "node-postgres".

gajus:
> Writing Date conversion manually is error-prone.

Please explain.

My reluctance for supporting `Date` is because `Date` instance is a mutable object.

Of course, one can simply cast it to number at the time it is provided, but then equally explicitly casting the value (in my opinion) better expresses the intent.

turuslan:
I've got your point.

`Object.freeze` doesn't work for both `Date` and `Buffer`, but `Date` has other way to pass value by explicit casting.

Thank you.

PR #59, opened 2019-05-15
fix: don't quote array type identifiers

mjm:
When used in a SELECT statement, either in WHERE clause or in the columns, the inclusion of double quotes causes PostgreSQL to not be able to parse the type name.

For example, this works:

```
=> select array[1,2,3]::int[];
  array  
---------
 {1,2,3}
(1 row)
```

but this doesn't:

```
=> select array[1,2,3]::"int"[];
ERROR:  type "int[]" does not exist
LINE 1: select array[1,2,3]::"int"[];
```

### DIFF
--- a/src/sqlFragmentFactories/createArraySqlFragment.js
+++ b/src/sqlFragmentFactories/createArraySqlFragment.js
@@ -4,16 +4,13 @@ import type {
   SqlFragmentType,
   ArraySqlTokenType
 } from '../types';
-import {
-  escapeIdentifier
-} from '../utilities';
 
 export default (token: ArraySqlTokenType, greatestParameterPosition: number): SqlFragmentType => {
   const values = [
     token.values
   ];
 
-  const sql = '$' + (greatestParameterPosition + 1) + '::' + escapeIdentifier(token.memberType) + '[]';
+  const sql = '$' + (greatestParameterPosition + 1) + '::' + token.memberType + '[]';
 
   return {
     sql,
--- a/test/slonik/sqlFragmentFactories/createArraySqlFragment.js
+++ b/test/slonik/sqlFragmentFactories/createArraySqlFragment.js
@@ -13,7 +13,7 @@ test('creates an empty array binding', (t) => {
     values: []
   }, 0);
 
-  t.assert(sqlFragment.sql === '$1::"int4"[]');
+  t.assert(sqlFragment.sql === '$1::int4[]');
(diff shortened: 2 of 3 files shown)


gajus:
For what it matters, if you use `int4` it will work as expected. I cannot recall the reason `::"int"[]` does not work but `::"int4"[]` does work.

mjm:
Ah, you're right. And it looks like other types work too, like "text". Well, then I guess it's up to you whether this is a good idea to merge. I found it surprising, and would certainly appreciate not having to be so particular about the types :) but I get it if you'd prefer to leave it as it is.

gajus:
> but I get it if you'd prefer to leave it as it is.

As a very least I would like to understand the underlying reason, which will explain if int4 should be preferred or if it is all the same. I've asked smarter people than me for feedback.

mjm:
Cool! If you end up leaving this as is, maybe a small addition to the README section for `sql.array` would help folks like me not get stuck on this :)

turuslan:
`int4` is an identifier ([pg_type.dat, line 76](https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_type.dat#L76), [gram.y, line 12563](https://github.com/postgres/postgres/blob/master/src/backend/parser/gram.y#L12563)) and can be double quoted.

`int` is a keyword with token `INT_P`, ([kwlist.h, line 213](https://github.com/postgres/postgres/blob/master/src/include/parser/kwlist.h#L213), [gram.y, line 649](https://github.com/postgres/postgres/blob/master/src/backend/parser/gram.y#L649)) which maps to `int4` type ([gram.y, line 12584](https://github.com/postgres/postgres/blob/master/src/backend/parser/gram.y#L12584)).
Identifiers may be double quoted but keywords can not.

Postgres has quite complex syntax for types ([gram.y, line 12463](https://github.com/postgres/postgres/blob/master/src/backend/parser/gram.y#L12463)).
Types can span multiple words and contain expressions like `character varying (20)`.
Only identifier name part of type can be double quoted.

It would be too difficult to implement behavior that will be compatible with PostgreSQL for `sql.array` type escaping.
It would be better to state limitations of `sql.array` and provide an example of array parameters using `sql.raw`.
Avoiding escaping type keywords may be an option, but just mentioning that some types will not work may be sufficient.

gajus:
Thank you @turuslan , @mjm for your input.

PR #61, opened 2019-05-17
Support TLS parameters in connection URI

### LINKED ISSUE #60: Unable to specify SSL cert and key
Given the following connection URL:

`postgres://user:pass@127.0.0.1?ssl=true&sslcert=/path/to/db.cert&sslkey=/path/to/db.key`

I get the following error:

`ConnectionError: connection requires a valid client certificate`

It appears Slonik does not recognize the `sslcert` and `sslkey` parameters. After some investigation, I believe this is related to https://github.com/iceddev/pg-connection-string/issues/8

I see a few options:

* Add support for these in https://github.com/iceddev/pg-connection-string
* Support [libpq environment variables](https://www.postgresql.org/docs/9.3/libpq-envars.html)
* Allow a config object. However, it appears that may not be preferred https://github.com/gajus/slonik/issues/45

hjr3:
The pg-connection-string dependency has been updated to support TLS
cert, key and ca parameters.

Fixes #60.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "iso8601-duration": "^1.2.0",
     "lodash": "^4.17.11",
     "pg": "^7.11.0",
-    "pg-connection-string": "^2.0.0",
+    "pg-connection-string": "github:hjr3/pg-connection-string#ssl-support",
     "pg-cursor": "^2.0.0",
     "pg-types": "^2.0.1",
     "postgres-interval": "^1.2.0",


hjr3:
@gajus I know dependencies can be a risk to any project. I will continue to try https://github.com/iceddev/pg-connection-string/pull/26 merged. In the meantime, I will commit to supporting https://github.com/hjr3/pg-connection-string for at least 12 months.

gajus:
Doesn't seem worthy of adding as a dependency given that the codebase is just 40 lines of code. Can we just internalize it as `parseConnectionString` function?

hjr3:
@gajus That was going to be my suggestion if you did not like the dependency. I was able to get a hold of the pg-connection-string maintainer and this may be merged in soon ™️ . Shall we wait 1-2 days to see if it is merged in before internalizing the code?

gajus:
> Shall we wait 1-2 days to see if it is merged in before internalizing the code?

Sounds good to me.

gajus:
`pg-connection-string` has since been updated to support this feature.

PR #74, opened 2019-07-17
feat: omit undefined values from sql.assignmentList generated sql code

n1ru4l:
This simplifies cases where you want to pass an object with possible undefined values trough too `sql.assignmentList`, without having to filter out `undefined` values via `delete object.key` or cloning.

### With `delete`
```js
const input = {
  foo: "bar",
  lel: undefined
};

if (input.hasOwnProperty("lel") && input.lel === undefined) {
  delete input.lel;
}

sql.assignmentList(input);
```

### Cloning
```js
const input = {
  foo: "bar",
  lel: undefined
};

const sanitizedInput = Object.fromEntries(
  Object.entries(input)
    .filter(([, value]) => value !== undefined)
);

sql.assignmentList(sanitizedInput);
```

## After:

```js
const input = {
  foo: "bar",
  lel: undefined
};

sql.assignmentList(input);
```

### DIFF
--- a/src/sqlFragmentFactories/createAssignmentListSqlFragment.js
+++ b/src/sqlFragmentFactories/createAssignmentListSqlFragment.js
@@ -20,6 +20,9 @@ export default (token: AssignmentListTokenType, greatestParameterPosition: numbe
 
   const sql = Object
     .entries(token.namedAssignment)
+    .filter(([, value]) => {
+      return value !== undefined;
+    })
     .map(([column, value]) => {
       if (isSqlToken(value)) {
         // $FlowFixMe
--- a/src/types.js
+++ b/src/types.js
@@ -317,8 +317,9 @@ export type ValueExpressionType =
   SqlTokenType |
   PrimitiveValueExpressionType;
 
+export type NamedAssignmentValueType = ValueExpressionType | void;
 export type NamedAssignmentType = {
-  +[key: string]: ValueExpressionType
+  +[key: string]: NamedAssignmentValueType
 };
 
 export type TaggedTemplateLiteralInvocationType = {|
--- a/test/slonik/templateTags/sql/assignmentList.js
+++ b/test/slonik/templateTags/sql/assignmentList.js
@@ -60,6 +60,21 @@ test('converts camel-case to snake-case', (t) => {
   });
 });
 
+test('omits undefined values', (t) => {
+  const query = sql`SELECT ${sql.assignmentList({
+    foo: 'baz',
+    lel: undefined
+  })}`;
+
+  t.deepEqual(query, {
+    sql: 'SELECT "foo" = $1',
+    type: SqlToken,
+    values: [
+      'baz'
+    ]
+  });
+});
+
 test('the resulting object is immutable', (t) => {
   const token = sql.assignmentList({foo: 'bar'});
 


gajus:
Set it explicitly to `null` or `omit` undefined properties using a utility.

`undefined` in such a scenario is often an indication of a bug – what is a real-world scenario where it would be intentional that values object value would have `undefined` value and it is not an oversight?

gajus:
This is very unlikely to be merged. Closing. But please continue discussion if I am missing something.

n1ru4l:
> `undefined` in such a scenario is often an indication of a bug – what is a real-world scenario where it would be intentional that values object value would have undefined value and it is not an oversight?

You have a query that updates `n`-fields of a table. 

```ts
class Foo {
  async updateIntegration({
    viewer,
    integrationId,
    data: { newTitle, newSecret, newWebhookUrl }
  }: {
    viewer: UserRecord;
    integrationId: number;
    data: { newTitle?: string; newWebhookUrl?: string; newSecret?: string };
  }) {
    if (!newTitle && !newWebhookUrl && !newSecret) {
      throw new ApolloError("Must update at least one value!");
    }
    const integration = await this.slonik.maybeOne(sql`
      UPDATE "integrations"
      SET ${sql.assignmentList({
        title: newTitle,
        webhookUrl: newSecret,
        secret: newWebhookUrl
      })}
      WHERE
        "userId" = ${viewer.id}
        AND "id" = ${integrationId}
      RETURNING ${integrationRecordFields()};
    `);

    return integration;
  }
}
```

gajus:
If you are feeling brave, then use [`omit`](https://lodash.com/docs/4.17.14#omit).

I would not allow that API. Property either must not exist or have a value that is intended to be updated/ inserted. `undefined` can easily be someone making a typo somewhere up the stream and that bug would obscured:

```js
updateIntegration(viewer, 1, {
  newTitle: input.newTitle,
  newSecret: input.newSecet
});

```

and just like that you block out all your users.

n1ru4l:
Well, I do not think that is very unlikely to happen (due to types), I can understand your concern.

I am curious how you would solve this issue? Having endpoints that allow patching multiple values is very common.

Something like the following:

```js
const updates = {};
    if (newTitle) {
      updates["title"] = newTitle;
    }
    if (newWebhookUrl) {
      updates["webhookUrl"] = newWebhookUrl;
    }
    if (newSecret) {
      updates["secret"] = newSecret;
    }

    if (Object.keys(updates).length === 0) {
      throw new ApolloError("Must update at least one value!");
    }
```
seems very verbose to me.

gajus:
> Having endpoints that allow patching multiple values is very common.

GraphQL Apollo server implementation does not populate parameter properties if they are unset, so you are safe in this particular setup (based on the name of the Error that you have used).

If you are conditionally bringing values from elsewhere, then IF guard for each of those is appropriate.

nodefish:
@gajus The assumption that `undefined` indicates a likely bug is quite incorrect. The distinction between `undefined` and `null` is in fact often used deliberately to express different concepts in JavaScript.

I appreciate the strictness of this project, but I think Knex got this one right.

gajus:
> between undefined and null is in fact often used deliberately to express different concepts in JavaScript.

That is correct. `undefined` means that variable is declared but that it is not assigned a value. `NULL` concept is shared between JavaScript and PostgreSQL (explicit no value).

> It does not make sense for sql`...` to output null for values that are undefined.

Slonik does no such thing. Slonik will throw an error if `undefined` is provided.

> It's an unduly opinionated decision and I would argue that it's less safe because it deliberately sets database fields to null without the programmer having passed null.

Slonik does no such thing.

nodefish:
I just encountered this situation and got null constraint errors from the db when passing an object with undefined values, so I thought that was Slonik's doing. I appreciate the correction. I'm not sure what happened (it's a complex endpoint so it's not immediately obvious).

I do nonetheless think undefined should be treated as OP suggests, despite my last objection being incorrect. It's very common to pass in a variable number of fields for an update, and it's quite safe to limit the fields using object destructuring, e.g. `const { name, email } = data;` in order to prevent the client from injecting additional columns into an update, but this doesn't mesh that nicely with Slonik (which btw. is excellent; thing tends to sound overly critical in GitHub "issues" for obvious reasons).

gajus:
> I do nonetheless think undefined should be treated as OP suggests

Please specify the expected behaviour.

OP is asking to ignore `undefined` properties. OP wrongly suggested that a scenario of unintentional undefined data is unlikely due to types. However, what OP is overlooking is that when you are working with remote data sources (APIs, user constructed input), you do not get run-time type safety.

Warning about properties that are present but have `undefined` value is a necessary precaution to avoid accidental loss of data.

If you share an example of code that is troubling you, I can suggest how I would would write such code and explain my reasons. If nothing else, this would at least put as on the same page.

nodefish:
I know how to solve it, but I appreciate the offer to help figure it out. 

I find this subjective:

>Warning about properties that are present but have undefined value is a necessary precaution to avoid accidental loss of data.

I don't at all believe that objects with properties that don't have values suggest anything accidental happened. Quite simply, very often you want to insert a variable set of fields determined by the client, and it's more practical to pass the object as-is (which most query builders I've used understand as deliberate).

I think both sides of this argument are hyperbolic (e.g. type systems preventing this, accidental loss of data) and really it's a design choice. If you don't feel it's aligned with Slonik's mission, then fair enough. But personally I think this pattern has value.

gajus:
Simple example of where this can backfire without any warning:

```js
type PersonType = {|
  +id: string,
  +name: string
|};

const getPerson = (): PersonType => {
  return require('http://get-person/');
};

const updatePerson = (person: PersonType) => {
  sql.assignmentList(person);
};

const main = () => {
  const person = await getPerson();
  
  updatePerson(person);
};

```

Types are going to be useless here if remote response changes.

There are _many_ things that can go wrong when working with external inputs if extensive runtime validation is not implemented. That is why it is important to practise defensive programming whenever an opportunity arrises. If user opts-to blindly pass input object to the query builder, then he has an option to `omit` all undefined properties using a helper function. I do not recommend it.

nodefish:
OP's suggestion is indeed definitely not a substitute for validation. But validation is often done in-place rather than having to construct another object with undefined values omitted. Typically, if just 1 value fails validation, the entire update operation is aborted. 

Ultimately it's really just a matter of eliding the specific bits of code whereby a new object is created (boilerplate).

gajus:
We could add this feature to https://github.com/gajus/slonik-utilities, though. Which is an opinionated tool belt. Raise an issue there if you would like to see it added. I would imagine that in most cases people would need this for `update`/ `upsert` only anyway.

Issue #42, opened 2019-03-06
Migration management with Slonik


strattadb:
Hi all! :wave:

I would love to manage my migrations using Slonik.

What do you think of adding migration management?

Great library, by the way!

gajus:
Can you define migration management requirements?

It is a very broad requirement at this point.

What benefits would it have to use Slonik over, e.g. https://github.com/sqitchers/sqitch?

Migration seems to be a separate issue from database driver, but I am happy to be proven wrong.

strattadb:
Of course. I was thinking more along the lines of what Knex.js does with migrations.

For example, we could do

```bash
slonik create-migration add_users_table
```
And it would create a file named something like `1551912556532_add_users_table.js` in a `migrations` folder.

And we could write the migrations using the `sql` function. For example:

```javascript
// 1551912556532_add_users_table.js

import { sql } from 'slonik'

export async function up(connection) {
  await connection.query(sql`
    CREATE TABLE users (
      id integer PRIMARY KEY,
      first_name varchar(64),
      last_name varchar(64),
      -- ...
    );
  `);
}

export async function down(connection) {
  await connection.query(sql`
    DROP TABLE users;
  `);
}
```

And then, we could do:

```bash
slonik migrate
slonik undomigration
# etc.
```

The main advantage would be that we could write our migrations with `sql`, which I think would be very nice.

What do you think?

makmm:
what are you all using for migrations right now? I really need a solution.

Sharaal:
I using `knex` for migrations, SQL Tags for query building and `pg` as database driver.
 `knex` solves the part of migrations more or less pretty well.

I think migrations, query building and database driver are completely different topics. On the one side its good to have all of these in one solution, especially if starting with these (so I also like tthe idea slonik provide migrations), but on the other side it's hard to have one thing fits in all use cases and like solutions which provide the possibilities to switch one or multiple layer of them if a project has different requirements (e.g. what if a project has a MySQL database).

Feel free to have a look at a blog article I wrote recently about this topic (if I'm allowed to link here): 
http://blog.sharaal.de/2019/03/12/knex-vs-alternatives.html

makmm:
@Sharaal Upon your suggestion, I used knex for migrations, but run into a case that I honestly don't understand how to solve in knex and much rather would learn the SQL way of doing it. Is there any other nice method for migrations?

Sharaal:
Whats the actual problem?

[...]

PR #105, opened 2019-10-03
Add section about migrations

karlhorky:
Ref #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -2110,4 +2110,4 @@ Slonik doesn't include migration management in the library - this is a philosoph
 
 > I like the idea of keeping Slonik simple and focused
 
-However, other libraries like [`node-pg-migrate`](https://github.com/salsita/node-pg-migrate) or [`@slonik/migrator`](https://github.com/mmkal/slonik-tools/tree/master/packages/migrator) can be good starting points.
+However, other libraries like [`node-pg-migrate`](https://github.com/salsita/node-pg-migrate) or [`@slonik/migrator`](https://www.npmjs.com/package/@slonik/migrator) can be good starting points.
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Note: Using this project does not require TypeScript or Flow. It is a regular ES
 * [Mapped errors](#error-handling).
 * [Atom plugin](#slonik-syntax-highlighting).
 * [ESLint plugin](https://github.com/gajus/eslint-plugin-sql).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2101,3 +2102,12 @@ To enable highlighting, you need to:
...
 For more information, refer to the [JavaScript Tagged Template Literal Grammar Extensions](https://github.com/gandm/language-babel#javascript-tagged-template-literal-grammar-extensions) documentation of `language-babel` package.
+
+<a name="slonik-migrations"></a>
+## Migrations
+
+Slonik doesn't include migration management in the library - this is a philosophical decision (as per [#42](https://github.com/gajus/slonik/issues/42#issuecomment-510146858)):
+
+> I like the idea of keeping Slonik simple and focused
+
+However, other libraries like [`node-pg-migrate`](https://github.com/salsita/node-pg-migrate) or [`@slonik/migrator`](https://github.com/mmkal/slonik-tools/tree/master/packages/migrator) can be good starting points.
```
mmkal: For @slonik/migrator, can this point at the npm package rather than github? i.e. https://npmjs.com/package/@slonik/migrator - that way if/when the code moves the link won't break. Plus, it puts the version number front and centre. I'd like people to know it's v0.x.x before they install it. I'd argue the same for the other package but I care more about mine.

karlhorky: Done, thanks for the feedback! Definitely matters more for monorepos.

gajus:
A section about migrations is simply not relevant to a database client library. It belongs at a higher level abstraction. This topic is easily discovered using /issues (and now through /pull).

Note, that this does not imply that I discourage use of any of these libraries. They are all fine approaches. If this becomes a re-occurring topic, I will dedicate a more thorough section in the docs.

lydell:
I don’t think people (like me) understand what a database client library is and is not. So it might be worth adding a “Migrations” section in the readme explaining things. Or an FAQ.

export-mike:
I'm really enjoying using slonik, so simple!

@gajus I could be learning how to use migrator by now, but I'm writing a comment to an issue. I say this is really relevant, digging into issues is just noisy and just time wasted 

Anyway keep up the great work your libraries are solid! 👍 and maybe reconsider adding a link in the README?

sebinsua:
Yes, could the migrations heading just say "this library intentionally doesn't handle migrations, but if you need a solution for this then read X".

PR #124, opened 2019-11-12
remove extra section in the Query Building section

paulovieira:
This section ("Difference from JSON.stringify") is repeated, probably from a copy-paste.


gajus:
Thank you

Issue #113, opened 2019-10-23
Allow Date object as a parameter/ result


gajus:
It has been [proposed](https://github.com/gajus/slonik/issues/94) that date types that describe timestamp should be parsed into Date object, and that when Date object is provided, it should be converted into a timestamp that PostgreSQL understands.

In general, I am opposed to using Date object at all, because of the mutable API. However, with regards to reading values from the database, I can see some value in returning `Date` as it provides a uniform data type for arbitrary accuracy timestamp values (in contrast, Slonik assumes millisecond accuracy).

With regards to accepting Date object as a parameter, would it require to simply translate Date to string  and use that as a value?

gajus:
cc @ivan-kleshnin @camflan

camflan:
If we're passing a `Date` instance to slonik, shouldn't we simply use `date.getTime()` to get the ms timestamp?

Re your concerns around mutability, I agree. I don't use Date directly, but instead via date-fns (and similar home grown fns) to avoid mutating the original Date. What are you using to deal with dates in an immutable fashion?

gajus:
> If we're passing a Date instance to slonik, shouldn't we simply use date.getTime() to get the ms timestamp?

Timestamp can be more accurate than milliseconds, though.

> What are you using to deal with dates in an immutable fashion?

date-fns

camflan:
> Timestamp can be more accurate than milliseconds, though.

Right, but if we're given a Date, then isn't ms is the most precise info we have?

> > What are you using to deal with dates in an immutable fashion?
> 
> date-fns

Cool - we're on the same page. 👍

gajus:
> Right, but if we're given a Date, then isn't ms is the most precise info we have?

Good point.

I am honestly not sure what is the best way to proceed.

If we were to return Date, and timestamp contains microseconds, then such implementation would truncate the data.

It might be reasonable compromise to make given that use of microseconds is pretty rare.

sgoll:
For reference, `node-postgres` itself converts date/time columns into `Date` objects and does the correct conversion when passed `Date` objects to forward them to the database:

https://node-postgres.com/features/types#date%20/%20timestamp%20/%20timestamptz

> Although PostgreSQL supports microseconds in dates, JavaScript only supports dates to the millisecond precision. Keep this in mind when you send dates to and from PostgreSQL from node: your milliseconds will be truncated when converting to a JavaScript date object even if they exist in the database. If you need to preserve them, I recommend using a custom type parser.

[...]

PR #130, opened 2019-11-25
implement sql.timestamp

rtbo:
As proposed in #113, provide sql.timestamp function
that takes a millisecond argument and  wraps a call
to `to_timestamp`.

If this is accepted, I will work on the following steps:
 - have timestamp usable in `sql.unnest` (and possibly other constructs).
 - provide `sql.date` that takes a Date argument and forwards to `sql.timestamp`.

### DIFF
--- a/src/factories/createSqlTag.js
+++ b/src/factories/createSqlTag.js
@@ -11,6 +11,7 @@ import type {
   SqlSqlTokenType,
   SqlTaggedTemplateType,
   SqlTokenType,
+  TimestampSqlTokenType,
   UnnestSqlTokenType,
   ValueExpressionType,
 } from '../types';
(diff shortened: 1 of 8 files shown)


gajus:
Doesn't accepting integer defeat the purpose of having `sql.timestamp` helper?

I thought that the whole point of doing this was so that developers could pass `Date` object.

rtbo:
Probably. The idea was to have millisecond timestamps usable in arrays or unnest, but I'm starting to figure out that it is not the proper way to do it.
I had the idea to do something like:
```js
const now = Date.now();
const arr = objArr.map(obj => [ obj.field1, obj.field2, sql.timestamp(now) ]);
await connection.query(sql`
    INSERT INTO table (field1, field2, updated_on)
    SELECT * FROM ${sql.unnest(arr, [ 'int4', 'int4', 'timestamp' ])}
`);
```
I understand now that it won't work and that I'd need to adapt the query (`SELECT to_timestamp(...) FROM ${sql.unnest(...)}`).

May be that's overkill anyway and I'd better pass a date string to unnest directly.

Issue #119, opened 2019-11-07
Pass custom log function


moltar:
It would be great to be able to replace Roarr with a custom logger.

Use case:

I'm currently using simple `debug`, and only in testing environment. This place very nicely with Jest.

Roarr output is JSON, and it cannot be parsed well when mixed with Jest output.

gajus:
No intent to support custom logger.

Roarr is used by many sub-packages of Slonik.

Consider replacing debug with Roarr: https://medium.com/free-code-camp/roarr-the-perfect-json-logger-node-js-and-browser-935180bda529

moltar:
Any suggestions as to how make it work mixed with other output?

E.g. as part of running tests.

gajus:
Requirements are unclear. Please clarify requirements and I am happy to help.

PR #158, opened 2020-02-29
feat(logging): allow passing roarr instance

mmkal:
related: https://github.com/gajus/slonik/issues/119

This will allow taking advantage of slonik's default logging, as well as packages like [slonik-interceptor-query-logging](https://github.com/gajus/slonik-interceptor-query-logging) to log full queries, at the same time as passing in existing application context (e.g. the incoming http headers for the request that triggered a sql query), by letting users pass in their own roarr instance into `createPool`.

Some notes/questions:

- ~to take advantage of this and pass in per-request context, it'd mean calling `createPool` once per request. It's a little hard to tell whether that will be an expensive operation or have any negative side-effects - [lines like this make me a little uneasy](https://github.com/gajus/slonik/blob/b29afbb6420f59e5d8584a9ea466921e2fe21c56/src/factories/createPool.js#L40). If it is, I could move the logger injection further down (maybe to `bindPool.js`, which looks like it only sets up some lazy methods), but it'd be a slightly bigger code change/API surface change, so I thought I'd check first.~
  - added a `withLogContext` method to `DatabasePoolType`, and `bindPool` implements it by simply recursively calling itself, with the logger argument modified using `parentLog.child`. This allows passing in a custom logger with application-level information to `createPool` (e.g. `roarr.child({ appName: 'booking-service', appVersion: 'v1.2.3' })`, but for individual requests, it's possible to call `myPool.withLogContext({ requestId: 'abc123' })` to get a cheap pool reference, with a new logger, without the potential side-effects of calling `createPool` many times.
- this isn't the full story in addressing the ask from @moltar in https://github.com/gajus/slonik/issues/119 - it would also be nice to be slightly less bound by roarr's design decisions. There's nothing I can see that's wrong with roarr at all, but I'm looking to integrate slonik into an existing project with a custom logging framework. We are fully in control of what gets printed to stdout, so wouldn't be comfortable turning on the potential `ROARR_LOG=true` firehose. For that reason, I may follow this up with a roarr PR that allows passing in a custom `onMessage` function to a roarr instance, to allow interop with custom logging libraries (which would hopefully cover @moltar's use-case of wanting special formatting for integration tests - shared by my team, too, in fact).

TODO:
- [x] unit tests
- [x] documentation of new parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -2137,6 +2137,31 @@ By default, Slonik logs only connection events, e.g. when connection is created,
 
 Query-level logging can be added using [`slonik-interceptor-query-logging`](https://github.com/gajus/slonik-interceptor-query-logging) interceptor.
 
+To include application-specific log context, pass in a `roarr` instance to the `createPool` parameters. You can also use `withLogContext` on an existing pool. For example in an http server request handler:
+
+```js
+import roarr from 'roarr';
+import { createPool, sql } from 'slonik';
+
+const appLogger = roarr.child({
+  appName: 'my-app',
+  appVersion: 'v1.2.3',
+});
+
+const pool = createPool(process.env.POSTGRES_CONNECTION_STRING, {
+  logger: appLogger,
+});
+
+export const handler = async (req, res) => {
+  const slonik = pool.withLogContext({
+    requestId: req.headers['x-request-id'],
+  });
+
+  const results = await slonik.any(sql`SELECT ...`);
+  res.status(200).send(results);
+};
+```
+
 <a name="slonik-debugging-capture-stack-trace"></a>
 ### Capture stack trace
 
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -3,7 +3,7 @@
 import Roarr from 'roarr';
 
 export const getLogger = (logger: typeof Roarr) => (logger || Roarr).child({
-  package: logger?.getContext().package || 'slonik',
+  package: 'slonik',
 });
 
 export default getLogger();
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -1,7 +1,9 @@
 // @flow
 
-import Logger from 'roarr';
+import Roarr from 'roarr';
 
-export default Logger.child({
-  package: 'slonik',
+export const getLogger = (logger: typeof Roarr) => (logger || Roarr).child({
+  package: logger?.getContext().package || 'slonik',
 });
+
+export default getLogger();
--- a/src/Logger.js
+++ b/src/Logger.js
@@ -2,7 +2,7 @@
 
 import Roarr from 'roarr';
 
-export const getLogger = (logger: typeof Roarr) => (logger || Roarr).child({
+export const getLogger = (logger?: typeof Roarr) => (logger || Roarr).child({
   package: 'slonik',
 });
 
--- a/src/binders/bindPool.js
+++ b/src/binders/bindPool.js
@@ -16,9 +16,8 @@ import {
 import {
   assertSqlSqlToken,
 } from '../assertions';
-import {bindPool} from '.';
 
-export default (
+const bindPool = (
   parentLog: LoggerType,
   pool: InternalDatabasePoolType,
   clientConfiguration: ClientConfigurationType,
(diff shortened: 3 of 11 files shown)


gajus:
Why not just use https://github.com/gajus/roarr#roarr-api-adopt ?

mmkal:
@gajus I looked at that but there are a few reasons:

- the deprecation notice. I know you covered that in your blog post, but it's still a legitimate concern. The fact that it's been deprecated for several years makes it more scary, not less.
- the fact that it's necessary to shim nodejs with yet another dependency (domain-parent), which is monkey-patching nodejs and can't be used in distributed applications, according to the warning.
- passing references explicitly is much easier to reason about and test.
- the change doesn't add much complexity to slonik at all

Another point is that with this change, it's still possible to use `adopt` exactly as you're suggesting. So if the above reasons don't apply to someone else, they aren't bound by them.

mmkal:
@gajus added some docs - it'd be great to get any feedback on what your concerns are, if any, with passing in a logger.

My aim is not to be prescriptive by saying it _must_ be done this way instead of with `.adopt`, just to enable it for teams like mine who would likely not be able to use `.adopt`. Might it help to also add an `adopt` example to the readme?

gajus:
I just don't see value in supporting this at this library level. I would rather consider extending [roarr](https://github.com/gajus/roarr) to allow global override of all log forwarding.

As mentioned in other discussion, now there is a proper way to implement `.adopt` without using domains. That would be the proper fix if this is the specific part that bothers users.

rintaun:
@gajus I am trying to add context to slonik log messages, e.g. the id of the http request that initiated particular events, and my first thought was that it would be nice if I could pass in a base roarr instance -- hence I ended up here.

I am able to accomplish what I'm trying to do with connection-level log messages using `adopt`, but not pool-level log messages, e.g. `client is checked out from the pool`, since `adopt` only allows static (i.e. call time) context, and I do not have the request id when the pool is initialized.

I understand it's quite pedantic, but can you think of any way to add that context to slonik's pool-level logging?

gajus:
I could be wrong, but can't you do:

```js
const pool = await log.adopt(() => {
  return createPool();
}, { /* context */ });

```

rintaun:
indeed you can!

here is my situation:

```javascript

const pool = await log.adopt(() => {
  return createPool();
}, { /* context */ });

app.get('/some/path', (req, res) => {
  return await log.adopt(() => {
    return pool.any(/* query */);
  }, { requestId: req.id }); // this
});
```

from that, I get something like the following output:

```
[2020-06-10T02:03:11.761Z] DEBUG (20) (@slonik): created a new client connection
poolId:    01EADZEWM6MWDS8PH2M4VSMC1N // no request ID ❌
processId: 42716
stats: 
  idleConnectionCount:  1
  totalConnectionCount: 2
  waitingRequestCount:  0

[2020-06-10T02:03:11.761Z] DEBUG (20) (@slonik): client is checked out from the pool
poolId:    01EADZEWM6MWDS8PH2M4VSMC1N // no request ID ❌
processId: 42716
stats: 
  idleConnectionCount:  1
  totalConnectionCount: 2
  waitingRequestCount:  0

[2020-06-10T02:03:11.765Z] DEBUG (20) (@slonik): executing query
requestId:    01EADZF3FB0WVE68YYM166WBBD // request ID ✅
poolId:       01EADZEWM6MWDS8PH2M4VSMC1N
connectionId: 01EADZF3GHQZQ4C0G8EQ2BEWA3
queryId:      01EADZF3GJNSMKFDAQJEE79ECZ
sql:          /* query */

[2020-06-10T02:03:11.773Z] DEBUG (20) (@slonik): query execution result
requestId:     01EADZF3FB0WVE68YYM166WBBD // request ID ✅
poolId:        01EADZEWM6MWDS8PH2M4VSMC1N
connectionId:  01EADZF3GHQZQ4C0G8EQ2BEWA3
queryId:       01EADZF3GJNSMKFDAQJEE79ECZ
executionTime: 11ms
rowCount:      4
```

because the callbacks for connect, acquire, etc. are created outside of any context i have access to dynamically. there may some shenanigans I could do with `async_hooks`, potentially, but I'm not familiar enough to really know...

PR #179, opened 2020-05-17
Allow sql.json to encode primitive values

gavacho:
I was working on something that took a passed unknown value (typed as `SerializableValueType`) and would write/read it from the database. I noticed that things worked fine for complex javascript objects/arrrays like `[{foo: 'bar'}]` but not when I passed in a string. This problem only occurred when I was using `sql.json` to encode the value. If I manually called `JSON.stringify` on the value, then it would be inserted into the database successfully.

I investigated and am providing this patch

### DIFF
--- a/src/sqlFragmentFactories/createJsonSqlFragment.js
+++ b/src/sqlFragmentFactories/createJsonSqlFragment.js
@@ -27,7 +27,7 @@ export default (token: JsonSqlTokenType, greatestParameterPosition: number): Sql
 
   // @todo Deep check Array.
   // eslint-disable-next-line no-negated-condition
-  } else if (!isPlainObject(token.value) && !Array.isArray(token.value)) {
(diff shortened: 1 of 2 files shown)


gajus:
I have mixed feelings about this. This could end up being abused by novice developers to insert primitives.

I am sensing that wanting to insert primitives with `json` is an edge case. There is nothing stopping you from checking if the type is primitive and falling back to using a different binding.

I am closing this for now, but if more people end up bumping in this issue, will reconsider. It could be enabled with a second parameter, i.e. `sql.json(payload, allowPrimitives: boolean)`

gavacho:
Ah, I guess it would be nice if the function signature of `sql.json` were updated to only allow object, array and null values then.

Thanks for considering

gajus:
Oh

gajus:
I didn't realise that the static types already permit this.

gajus:
Merged.

PR #183, opened 2020-06-09
build on prepare

felamaslen:
If installing a specific revision of this library as a dependency, the package isn't built.

This should fix this so it's possible to do something like:

`yarn add https://github.com/gajus/slonik#some-git-rev-hash`


gajus:
Use NPM. This repository is not maintained to be installed via git.

PR #190, opened 2020-07-09
fix: apply type parsers to streamed rows

aravindanve:


aravindanve:
Maybe don't merge this yet, I may have found a simpler fix after digging around node-postgres
https://github.com/brianc/node-postgres/issues/2270#issuecomment-656329509

PR #193, opened 2020-07-12
Preserve ssl configuration in connection string

hjr3:
A connection string may contain ssl related query parameters. For
example: 'sslcert=/path/tocert'. This will add the contents of the
certificate to { ssl: { cert: '...' } }.

We continue to default rejectUnauthorized if ssl is turned on, but now
we respect all ssl configuration settings specified in the connection
string.

### DIFF
--- a/src/factories/createPoolConfiguration.js
+++ b/src/factories/createPoolConfiguration.js
@@ -29,6 +29,7 @@ export default (connectionUri: string, clientConfiguration: ClientConfigurationT
   if (poolConfiguration.ssl) {
     poolConfiguration.ssl = {
       rejectUnauthorized: false,
+      ...poolConfiguration.ssl,
     };
   }
 


hjr3:
Here is what was going on:

pg-connection-string was returning an object that looked like:

```
[Object: null prototype] {
  ssl: {
    cert: '-----BEGIN CERTIFICATE-----\n' +
      'hidden\n' +
      '-----END CERTIFICATE-----\n',
    key: '-----BEGIN RSA PRIVATE KEY-----\n' +
      'hidden\n' +
      '-----END RSA PRIVATE KEY-----\n'
  },
  sslcert: '/path/to/cert',
  sslkey: '/path/to/key',
  user: 'user',
  password: 'password',
  port: null,
  host: 'localhost',
  database: 'spool'
}
```

but the logic in slonik was transforming it into

```
[Object: null prototype] {
  ssl: {
     rejectUnauthorized: false
  },
  sslcert: '/path/to/cert',
  sslkey: '/path/to/key',
  user: 'user',
  password: 'password',
  port: null,
  host: 'localhost',
  database: 'spool'
}
```

PR #209, opened 2020-10-19
BREAKING CHANGES: Flow -> Typescript

### LINKED ISSUE #195: Rewrite in TypeScript
I will be tackling this in the coming weeks.

Bear in mind that my familiarity with TS is limited. Therefore, expect a few rough iterations.

Anyone who is well versed in TS is more than welcome to jump in an lead.

This will be included in the next major release.

mmkal:
Fixes #195 

I aimed to change as little as possible to get things building, linting and testing successfully. There are a lot of potential follow-ups, but for now I wanted to make sure I wasn't affecting any runtime behaviour, so almost all changes are type assertions, and typescript/eslint disable directive comments. i.e. code that gets stripped away at compile time.

It's still using babel to transpile ts -> js. Typescript is used for type-checking of the whole project at lint time, and for ts->d.ts declaration generation at build time. It'd probably be a good idea to simplify by dropping babel entirely, but right now [inline-loops.macro](https://npmjs.com/package/inline-loops.macro) and babel-plugin-transform-export-default-name are being used which are babel-only. I'd recommend dropping inline-loops.macro in favour of plain-old `array.map(...)`. On my machine `.map` was marginally faster, but [they were nearly identical](https://gist.github.com/mmkal/2f242f4aa4f36181ca411dfd99496ea4). For babel-plugin-transform-export-default-name, it's an awesome plugin, but to reduce the complexity of maintaining two build tools, it might make sense to use a lint rule which prefers named exports over default exports for the next major.

The lint config is not as clean and minimal as before, because `eslint-config-canonical` doesn't know about typescript (yet). I added `@typescript-eslint/eslint-plugin` which has mostly equivalent rules (e.g. `flow/no-weak-types` -> `@typescript-eslint/no-explicit-any`. There are a few more `// eslint-disable`s that aren't related to typescript, because I didn't realise until halfway through that there are a few warnings that were already happening in master. I added `--max-warnings 0` to make sure warnings don't sneak in again but happy to take this out if necessary.

I commented out the `semantic-release` call in .travis.yml - it probably makes sense to release this manually. Or, this could be merged into a `v23` branch rather than master - @gajus up to you!

Some type-level tests were added in `test/types.ts` using [expect-type](https://npmjs.com/package/expect-type). More can be added here to make sure the types are working the way they're expected to. It'll probably be worth some explicit tests here that make sure user-facing methods don't return `any`, etc.

There are some formatting changes that came from the tool I used to convert flow to typescript, which uses prettier under the hood. There aren't too many, but if we want to eliminate them entirely, I'd recommend adding prettier to the lint step in master first (the codebase is ~95% in line with how prettier wants it to be anyway, so it shouldn't be too controversial). That way prettier can be run here too, and the only changes will be meaningful ones. Not a huge issue though, the formatting changes are easy enough to identify at a glance.

Follow-up recommendations:
- turn on `noImplicitAny` in tsconfig. I'd recommend doing this before a release. Left out here to keep the change as small as possible, but since typescript doesn't have usage-inference, this is pretty important to go in before functional changes start being made in typescript
- turn on `strictNullChecks` in tsconfig. Ditto
- turn on `strict` in tsconfig. The above two are the most important but there are a few more safety checks typescript can provide going forward
- drop inline-loops.macro and babel (see above)
- turn on `@typescript-eslint/prefer-ts-expect-error`. All the `$FlowFixMe`s I converted to `// @ts-ignore`. But `// @ts-expect-error` is better because it flags when the suppression isn't necessary
- then, do an audit of all the `@typescript-eslint/no-explicit-any` and `@ ts-expect-error` instances and see if they can be improved via refactoring. There will likely be _more_ offenders after turning on `noImplicitAny`.
- remove `// @flow` annotations. These are currently doing no harm. I left them in to make the diff easy to read. If I'd taken them out, every file would have changed, making it harder to spot the differences which actually meant something.
- update readme to remove flow references
- unpublish `@types/slonik` from DefinitelyTyped

Each of those is less of a pure language-conversion task, so IMO best left to separate PRs, even if the changes are released together.


gajus:
> I commented out the semantic-release call in .travis.yml - it probably makes sense to release this manually. Or, this could be merged into a v23 branch rather than master - @gajus up to you!

We can keep semantic-release, just need to make sure that the merge commit says "BREAKING CHANGES:"

gajus:
> The lint config is not as clean and minimal as before, because eslint-config-canonical doesn't know about typescript (yet).

Are you working on this already? I had it on my TODO list for a while, but happy to accept a PR too.

gajus:
Based on everything I read, the only question I have – do we want to release this now, do we want to merge this into a branch, or do we want to continue building this PR up? If you are Okay with leading these changes, then latter is preferable, but I can also swoop in to help.

mmkal:
> Based on everything I read, the only question I have – do we want to release this now, do we want to merge this into a branch, or do we want to continue building this PR up? If you are Okay with leading these changes, then latter is preferable, but I can also swoop in to help.

I'd rather merge into a branch if that's ok. I'm happy to work on most of the follow-ups I listed immediately (like today), but a few of them may require more context on slonik's design/implementation. Happy to try them all out but I think keeping things somewhat separated-out will make the change easier to follow, and easier to collaborate on without merge conflicts etc.

A branch would also solve the semantic-release problem - each PR could be squash-merged into a `v23` branch then a regular merge into master would trigger a release. I've just added `BREAKING CHANGES: ` to the PR title so if it's squash-and-merged in, the commit message will include it too.

If you could create a branch off master in this repo, I can update the base for this PR.

> > The lint config is not as clean and minimal as before, because eslint-config-canonical doesn't know about typescript (yet).
> 
> Are you working on this already? I had it on my TODO list for a while, but happy to accept a PR too.

No I'm not - and I don't think it's essential for the conversion to typescript. So I'd rather get this out and then see if I have time later - although lint configs are a lot about preference so it might make sense for your team to decide which typescript rules you want to include (I noticed you're the author of eslint-config-canonical too!).

mmkal:
> > I commented out the semantic-release call in .travis.yml - it probably makes sense to release this manually. Or, this could be merged into a v23 branch rather than master - @gajus up to you!
> 
> We can keep semantic-release, just need to make sure that the merge commit says "BREAKING CHANGES:" This will cause major version bump, which is the goal here anyway (since otherwise people's built systems that rely on Flow type checking will fail)

Just another note re semantic release. You're right that this should be a major bump because of the removal of flow types, but it'd be good to avoid another major bump immediately afterwards. There are a couple of places where the types can be made stricter than what they are in this change, which would be a breaking change if we released as-is. Better to put all the breaking changes in the 23.0.0 release, I think.

src/utilities/isPrimitiveValueExpression.ts
```diff
@@ -1,5 +1,6 @@
 // @flow
 
-export default (maybe: *): boolean %checks => {
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export default (maybe: any): boolean => {
```
danielrearden, line 4: We could make this a type guard to provide equivalent functionality
```suggestion
export default (maybe: any): maybe is string | number | boolean | null => {
```

gajus:
We now have https://github.com/gajus/slonik/tree/typescript branch.

mmkal:
Thanks for merging - @gajus, what are you thoughts on dropping inline-loops.macro and babel-plugin-export-default-name?

gajus:
Is the suggestion simply because it enables us to drop Babel? If that's the only reason, then I am reluctant to do it.

mmkal:
Yes. Having two separate build/transpilation tools seems overkill. Maintaining babelrc and tsconfig, making sure they merge files into the same directory, the various `npm run` scripts etc.

There's no typescript way of adding default export names, but a lint rule could make sure all exports have names anyway, with less magic.

I've worked on projects where both are used, and it works, but there's a maintenance overhead. Up to you.

gajus:
I am assuming nyc can be made to work with TS without Babel?

If so, I am Okay to remove Babel.

PR #210, opened 2020-10-19
TypeScript: enable strict flags

mmkal:
Note: there are a few spots where things are too "dynamic" to be easily trackable by the type system, I resolved a few by using `any` explicitly. IMO this isn't a big issue since they're all internal and hopefully the code paths are well-tested. But I'd advise dealing with them more rigorously as refactors go in over time.

Next, I think it'd make sense to pull in the tests from the `@types/slonik` package, since they validate how current typescript users work with slonik: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/slonik/slonik-tests.ts. But no point doing that before these two flags have been enabled because they make quite a big difference to the typescript used.


mmkal:
@gajus again the individual commits here wouldn't pass CI so makes sense to squash-and-merge into the `typescript` branch if the latest one looks good.

PR #211, opened 2020-10-20
TypeScript: tests from definitely typed

mmkal:
Related: #209 
Part of #195

This adds tests defined for the types in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/slonik/slonik-tests.ts - this is the API existing typescript users are working with - so after this, the only breaking changes for typescript users will be removal of some helpers like the interceptor presets that should already have gone from the definitions.

The tests are in `test/dts.ts`, GitHub may collapse it on some browsers since it's fairly long, but it's (almost) a straight copy-paste.

Exports some new methods to make them pass. Some modifications made for breaking changes that weren't ported over there.

Uses expect-type for testing since definitely typed uses tslint/dtslint which is deprecated.

This also replaces most `any` instances in `src/types.ts` with generics.

After this:
- switch from babel->tsc for building and @babel/register->ts-node for tests (I have a local branch for this on top of this one)
- convert default exports to named exports to compensate for [babel plugins/equivalents](https://github.com/gajus/babel-plugin-transform-export-default-name) not existing in typescript (likewise, this is ready)
- remove `// @flow` annotations and switch `// @ts-ignore`s to `// @ts-expect-error`s (mostly just find and replace)

At that point I think everything should be read to re-enable `semantic-release` in .travis.yml and merge `typescript` into `master`! @gajus I should be able to do all that today.


danielrearden:
@mmkal This looks amazing 😍

README.md
```diff
@@ -192,7 +191,7 @@ const getFooIdByBar = async (connection: DatabaseConnectionType, bar: string): P
 
 ```js
 const getFooIdByBar = (connection: DatabaseConnectionType, bar: string): Promise<DatabaseRecordIdType> => {
-  return connection.oneFirst(sql`
+  return connection.oneFirst<DatabaseRecordType>(sql`
```
gajus, line 194: What's the reason all the first types ask for `DatabaseRecordType`?

This should be a primitive value, e.g.

```js
connection.oneFirst<number>(sql``);
```

mmkal: Good point. The simple reason is that this is how it works in `@types/slonik`. The slightly more complicated reason is that the generic should really be on the tag, not the function call:

```typescript
connection.oneFirst(sql<DatabaseRecordType>`select id from foo`)
```

And there it _should_ be `DatabaseRecordType`, because the template `select id from foo` represents a pg return value looking like `{ rows: [{ id: '...' }] }`. So that query's typearg should be the full record. But I'll see if it's possible to switch up the `oneFirst`, `maybeOneFirst`, `anyFirst` and `manyFirst` type args to be more intuitive anyway. Do we care about breaking changes for `@types/slonik` users?

mmkal: @gajus updated - it's now

```typescript
connection.oneFirst<number>(sql``) // resolves to type `number`
connection.maybeOneFirst<number>(sql``) // resolves to type `number | null`
connection.anyFirst<number>(sql``) // resolves to type `number[]`
connection.manyFirst<number>(sql``) // resolves to type `number[]`
```

per your suggestion.

gajus:
🥳

PR #212, opened 2020-10-20
TypeScript: drop babel

mmkal:
Related: #209 
Part of #195

This removes 10 (!) babel-related dev dependencies, and switches over build and test commands to use typescript only.

Some type changes were needed because `inline-loops.macro` has pretty loose types, and was letting us get away with more hand-waving than `Array.prototype.map` does.

This loses the benefit of automatically naming anonymous functions (e.g. default exports), so in a follow-up PR I'm planning on switching to named exports. Lastly, I'll audit unnecessary comment directives.

`nyc` worked fine for me locally with `ts-node`, but let's see if coveralls is still able to pick up the report.


gajus:
This is great

PR #213, opened 2020-10-20
TypeScript: convert default exports to named

mmkal:
Related: #209 
Part of #195

Since switching from babel to typescript for building in #212, we can no longer use a plugin to give default-exported functions names as a transpilation step, which would make call stacks worse (default-exported functions appear anonymous). I used a dumb script to do the conversion, which probably would fall over for projects that weren't as consistent with naming conventions, etc.: https://gist.github.com/mmkal/e2d8434c9f7061a173bb32216af87a5a

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,6 +15,7 @@
     "linebreak-style": 0,
     // https://github.com/typescript-eslint/typescript-eslint/issues/1624#issuecomment-589731039
     "import/no-unresolved": 0,
+    "import/no-default-export": 1,
     "no-extra-parens": 0,
     "no-unused-vars": 0,
     "@typescript-eslint/no-unused-vars": 1,
@@ -47,6 +48,7 @@
       "files": "*.d.ts",
       "rules": {
         "filenames/match-regex": 0,
+        "import/no-default-export": 0,
         "import/unambiguous": 0
       }
     }
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "semantic-release": "^17.2.1",
     "sinon": "^9.2.0",
     "ts-node": "^9.0.0",
-    "typescript": "^4.0.3"
+    "typescript": "^4.1.0-beta"
   },
   "engines": {
     "node": ">=10.0"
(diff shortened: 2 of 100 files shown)


gajus:
What's the problem with default export?

As long as you name the export, it would work just as fine, i.e.

```js
const foo = () => {};

export default foo;

```

In terms of stack traces, the export will retain name.

mmkal:
The main problem is that it's hard to enforce that developers will actually do that. It's easier/more natural to do:

```js
export default () => {};
```

Which is what most modules in the codebase are currently doing. This is presumably the reason for babel-plugin-default-export-name existing in the first place. By banning default export, it _guarantees_ every exported function has a name, both now and in future. It also avoids the overhead of declaring the const in one place and having to export it separately.

Named exports are also recommended in the canonical TypeScript book (possibly because of the lack of plugins!): https://basarat.gitbook.io/typescript/main-1/defaultisbad

papb:
By the way, default exports is also a source of headache when integrating with CommonJS syntax

gajus:
> By the way, default exports is also a source of headache when integrating with CommonJS syntax

How so?

mmkal:
If you're using _pure_ commonjs, it can require doing slightly odd things like

```js
const myModule = require('./my-module')

myModule.default(...)
```

Or sometimes (and it's not always clear when to use which - it often depends on babel/tsconfig settings):

```js
const myModule = require('./my-module')

myModule(...)
```

@papb may have more examples, but in slonik's case that's not a major issue, since all the modules with default exports are internal.

But the anonymous function problem will be user facing unless anonymous default exports are converted to named functions. It's more conventional (at least in typescript projects I've worked on) to just use `export const foo` and cut out the middleman needed in the `const foo = ...; export default foo`.

Developer tools deal with named exports better too (e.g. VSCode) - possibly just because more work has been put into making the experience smooth, again possibly because of it being a typescript convention.

P.S. I believe all that's left after this PR is the flow annotation cleanup, so would be great to move forward one way or another!

gajus:
I am not a fan of this. Perhaps this is a getting-used type of thing, but default exports clearly signal that, well, this is the default export.

I have raised an issue suggesting a style rule for this https://github.com/benmosher/eslint-plugin-import/issues/1927

However, meanwhile, I don't see reason for this to be a blocker.

PR #214, opened 2020-10-21
TypeScript: cleanup comment directives

### LINKED ISSUE #195: Rewrite in TypeScript

mmkal:
Related: #209
Fixes #195

This is the last part of the TypeScript conversion. At this point it's ready to go in, but it may be worth rebase-squashing some of the commits, corresponding to each PR - that will make the history easier to follow.

There will always be improvements that can be made, like getting rid of `any`s and `// @ts-expect-error` directives, but I think there are now fewer of those than there were with Flow - so no reason to block on the remaining ones.

Happy to help out if there are any problems, but I don't have plans on doing any more after this, other than unpublishing `@types/slonik` when the next version is published.


danielrearden:
@mmkal Do you have any thoughts around how we could improve the typings for all the "first" methods? As a library consumer, I find it disappointing to have to write `<{ some_column_name: string }>` instead of `<string>` for these methods.

As I understand it, the only reason we pass the method's type variable to `QueryMethodParams` is so that it can be passed all the way down to `TaggedTemplateLiteralInvocationType`, and it's not even used explicitly by that type. I'm assuming the only purpose of the type variable is to be used for type inference by other libraries like `slonik-tools`.

This would be a significant breaking change, so it'd be nice to merge it into the `typescript` branch as well.

mmkal:
@danielrearden agreed - @gajus made the same point and I made that breaking change in one of the previous PRs: https://github.com/gajus/slonik/pull/211#discussion_r508729395

Paraphrasing here, the generics now work like this:

```typescript
connection.oneFirst<number>(sql`select id from foo where name = 'bar'`) // resolves to type `number`
connection.maybeOneFirst<number>(sql`select id from foo where name = 'bar'`) // resolves to type `number | null`
connection.anyFirst<number>(sql`select id from foo where name = 'bar'`) // resolves to type `number[]`
connection.manyFirst<number>(sql`select id from foo where name = 'bar'`) // resolves to type `number[]`
```

Basically now the type arg for each method (e.g. `.oneFirst`) is separate from the type arg for the `sql` tag. You could rewrite the above using a generic on the tag instead:

```typescript
connection.oneFirst(sql<{ id: number }>`select id from foo where name = 'bar'`) // resolves to type `number`
connection.maybeOneFirst(sql<{ id: number }>`select id from foo where name = 'bar'`) // resolves to type `number | null`
connection.anyFirst(sql<{ id: number }>`select id from foo where name = 'bar'`) // resolves to type `number[]`
connection.manyFirst(sql<{ id: number }>`select id from foo where name = 'bar'`) // resolves to type `number[]`
```

This makes sense, because the query itself doesn't know what method it will be passed to, so it needs to represent the actual sql query. In production applications, it'll often be instantiated separately from where it's used. e.g.:

```typescript
const idQuery = sql<{ id: number }>`select id from foo where name = 'bar'`

connection.oneFirst(idQuery) // resolves to type `number`
connection.maybeOneFirst(idQuery) // resolves to type `number | null`
connection.anyFirst(idQuery) // resolves to type `number[]`
connection.manyFirst(idQuery) // resolves to type `number[]`

// idQuery can also be used with the non-"first" functions:
connection.many(idQuery) // resolves to Array<{ id: number }>
```

[@slonik/typegen](https://npmjs.com/package/@slonik/typegen) is basically just automating and taking the possibility for typos out of that process.

There are a bunch of type-tests for this here: https://github.com/gajus/slonik/blob/6addcbfbce2264a0a7f7425a700fd69d0451a8e8/test/types.ts#L26-L71

danielrearden:
@mmkal Sorry, totally missed that comment and the subsequent change. Awesome 🤩

PR #234, opened 2020-11-22
feat: include pg message in errors

### LINKED ISSUE #233: Include message from originalError message in wrapped error
Related: #50

## Expected Behavior
Postgres error messages are surfaced to slonik errors and easily readable.

## Current Behavior
When errors are thrown (unique constraints, non-null etc.), the top line of the error message consists only of the generic message

This makes problems harder to track down/debug. The information is there, but when using tools that show only error messages (rather than full call stacks with fully serialised error objects), all you see is something like `Query violates a unique integrity constraint`.

Tools that work this way:
- sentry - it can show extra details as tags, but the error message is usually bubbled up to an issue title. This affects grouping. Currently, errors of this type all get grouped together, even for different columns.
- some command line tools. To keep the UI clean, some of these show the error message only, so it's necessary to attach a debugger, or add additional log statements and re-run queries to get at the `originalError` value. It's sometimes not possible to do this, and can be a frustrating task for intermittent errors.
- lots of organisation-specific error loggers/reporters. Ideally they'd always capture all the relevant information, but when they don't, it'd still be very useful to be able to see what's wrong at a glance.

In general, it's super-valuable for slonik to be mapping from error codes to error "types". But the postgres team has presumably spent a lot of time designing the error messages so they're clear and human readable, so the messages like `Query violates a unique integrity constraint` aren't helping much. So it'd probably even be ok to use `originalError.message` directly too.

## Possible Solution
The `SlonikError` class could have something like a `getMessage` method, which `originalError` is passed into, as well as an additional string. Then `this.message = ....` could become `this.message = this.getMessage(originalError, 'Query violates a unique integrity constraint')`.

## Steps to Reproduce

```js
let slonik = createPool(...)
await slonik.query(sql`create table t1(id int unique, name text unique`)
await slonik.query(sql`insert into t1(1, 'a')`)
await slonik.query(sql`insert into t1(2, 'a')`).catch(e => console.log(e.message))
```

I'd be happy to take a stab at a small refactor if you'd be open to it?

mmkal:
Fixes #233

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -3,7 +3,18 @@
 import ExtendableError from 'es6-error';
 
 export class SlonikError extends ExtendableError {
-  originalError!: Error
+}
+
+export class WrappedPGError extends SlonikError {
+  readonly message!: string;
+
+  readonly originalError: Error
+
+  constructor (originalError: Error, message: string) {
+    super(`${message}: ${originalError.message}`);
+
+    this.originalError = originalError;
+  }
 }
 
 export class InvalidConfigurationError extends SlonikError {}
@@ -14,51 +25,42 @@ export class UnexpectedStateError extends SlonikError {}
 
 export class ConnectionError extends SlonikError {}
 
-export class StatementCancelledError extends SlonikError {
-  constructor (error: Error) {
-    super();
-
-    this.originalError = error;
-    this.message = 'Statement has been cancelled.';
+export class StatementCancelledError extends WrappedPGError {
+  constructor (error: Error, message = 'Statement has been cancelled') {
+    super(error, message);
   }
 }
 
 export class StatementTimeoutError extends StatementCancelledError {
   constructor (error: Error) {
-    super(error);
-
-    this.message = 'Statement has been cancelled due to a statement_timeout.';
+    super(error, 'Statement has been cancelled due to a statement_timeout');
   }
 }
 
-export class BackendTerminatedError extends SlonikError {
+export class BackendTerminatedError extends WrappedPGError {
   constructor (error: Error) {
-    super();
-
-    this.originalError = error;
-    this.message = 'Backend has been terminated.';
+    super(error, 'Backend has been terminated');
   }
 }
 
 export class NotFoundError extends SlonikError {
   constructor () {
-    super('Resource not found.');
+    super('Resource not found');
   }
 }
 
 export class DataIntegrityError extends SlonikError {
   constructor () {
-    super('Query returns an unexpected result.');
+    super('Query returns an unexpected result');
   }
 }
 
-export class IntegrityConstraintViolationError extends SlonikError {
+export class IntegrityConstraintViolationError extends WrappedPGError {
   constraint: string;
 
-  constructor (error: Error, constraint: string) {
-    super();
+  constructor (error: Error, constraint: string, message = 'Query violates an integrity constraint') {
+    super(error, message);
 
-    this.originalError = error;
     this.constraint = constraint;
   }
 }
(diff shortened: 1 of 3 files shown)


test/slonik/integration.ts
```diff
@@ -997,6 +997,23 @@ test('throw error with notices', async (t) => {
...
+
+  await pool.query(createPerson123);
+
+  const error = await t.throwsAsync(async () => {
+    return pool.query(createPerson123);
+  });
+
+  t.regex(
+    error.message,
+    /^Query violates a unique integrity constraint: duplicate key value violates unique constraint "person_pkey"$/,
```
mmkal, line 1015: Any thoughts on getting rid of the "Query violates a unique integrity constraint" part completely? We'd still get the type safety by having the `UniqueIntegrityConstraintViolationError ` class, and the other part conveys the same information, but a bit more helpfully/specifically.

src/errors.ts
```diff
@@ -3,7 +3,18 @@
...
-  originalError!: Error
+}
+
+export class WrappedPGError extends SlonikError {
+  readonly message!: string;
+
+  readonly originalError: Error
+
+  constructor (originalError: Error, message: string) {
+    super(`${message}: ${originalError.message}`);
```
gajus, line 14: What's the reason for wanting to include the original error message?

test/slonik/integration.ts
gajus: I am of the inverse opinion that we shouldn't include: `duplicate key value violates unique constraint "person_pkey"` part.

I am generally of opinion that error messages should be static, and that details such as "which constraint" belong as properties of the error.

src/errors.ts
```diff
@@ -68,36 +70,24 @@ export class IntegrityConstraintViolationError extends SlonikError {
...
+    super(error, constraint, 'Query violates a unique integrity constraint');
   }
 }
 
 export class CheckIntegrityConstraintViolationError extends IntegrityConstraintViolationError {
   constructor (error: Error, constraint: string) {
-    super(error, constraint);
-
-    this.originalError = error;
-    this.message = 'Query violates a check integrity constraint.';
```
gajus, line 101: It is generally a bad idea to make minor stylistic changes such as removing a dot from the end of a sentence, because these are subjective and do not directly relate to the issue that the PR is trying to address.

gajus:
I've set a reminder to comeback to this PR tomorrow.

My main concern is that there is no guarantees that `pg` will remain the underlying driver for Slonik. Therefore, exposing errors like this creates unnecessary tight coupling and (future) breaking change.

src/errors.ts
mmkal: I removed it because when concatenated with the original message and the colon, it looks strange: `Query violates unique constraint.: duplicate key value violates unique constraint "person_pkey"`.

But I agree with you about keeping changes minimal, so I'll get rid of the colon instead.

mmkal:
There's always a tradeoff between error specificity and stability. I think of Error messages as _mostly_ for humans. So they should be short, predictable and understandable, but also deliver an appropriate amount of information. In general I find postgres does a good job finding that balance.

Right now slonik abstracts away the slightly more precise postgres message. To take the example from the linked issue:

```js
let slonik = createPool(...)
await slonik.query(sql`create table t1(id int unique, name text unique)`)
await slonik.query(sql`insert into t1 values(1, 'a')`)
await slonik.query(sql`insert into t1 values(2, 'a')`).catch(e => console.log(e.message))
```

In the above code, we'll only ever see "Query violates a unique integrity constraint.", whereas with the original message, we'd see "duplicate key value violates unique constraint "t1_name_key". With the original error message, you can see at a glance which table and column is the root of the problem. Of course, `e.originalError` already has that information, but `e.originalError` is often not present:

- At 3am when a developer gets paged. The alert says "Query violates a unique integrity constraint.". Where do they start looking?
- At 3:05am when they look at Sentry. All of the "Query violates a unique integrity constraint." errors are grouped together, including totally unrelated ones going back days, weeks or months, because Sentry uses a fairly naive algorithm for grouping. The increasingly panicked developer now doesn't know when this issue started happening and needs to dive deep into logs to (hopefully) get a better sense.
- When using command-line tools like [ts-command-line](https://npmjs.com/package/@rushstack/ts-command-line). By default, these only show the error message (no stack, and no attached properties). If you're lucky, when you rerun the command with some extra logging, the error will occur again. If it was a race condition or intermittent bug, you might be out of luck.
- In code like the above, where something is non-critical, so the team write an overly-permissive `.catch` and only log the error message, not all the properties associated (in this case, they _shouldn't_ do that. But when they do, the only clue to go on will be the generic error).

As for coupling - you're right, there's somewhat increased coupling - but isn't to postgres itself, rather than `pg`?  I believe `pg` is just surfacing postgres error messages. I wouldn't necessarily advocate for using the original error message unless postgres was such a solid, battle-tested system with very well-designed error messages. And if properly documented, a driver change wouldn't need to be considered breaking - using `Error` subclasses enables `if (e instanceof UniqueConstraintError) { ... }`, and `.toString()` will include the class name in the stack/message, so if the convention was to rely on that, it'd still be possible to correlate errors from another driver if desired.

mmkal:
@gajus bump on this. If you want to keep the errors as they are, what if I made it configurable? Right now there's no way as a user for me to switch to the original error message. I could add a function `WrappedPGError.getMessage(originalError, message)` which just returns `message` here, but a downstream user would be able to override the implementation of `getMessage` to return `originalError.message`, or `originalError.code + originalError.message`, or whatever they wanted, at their own risk.

Open to that idea?

gajus:
Thank you for bumping the conversation.

Your arguments make sense.

I have pushed the update.

PR #241, opened 2020-12-17
fix: remove deepFreeze from sql tag construction (fixes #227)

### LINKED ISSUE #227: sql.json recursively mutates object with deepFreeze
## Background
I recently ran into a particularly difficult to debug situation. 

I have an API Client which uses the http2 library to queue and perform requests to an API server. Attached to this client and each request is a `PromiseToken`, which acts as a cancellation token class.

After processing GBs of data, I found my code would ultimately fail with Cannot assign to read only property 'waitParent' of object '#<PromiseToken>'. I searched through all of my compiled code for all interactions with promiseToken and even all calls to `Object.freeze`, `Object.seal`, `Object.defineProperty`, and `Object.defineProperties`. No interactions or calls that would affect this. I even started digging in to see if it was an issue with Node.

I eventually set up hooks for all of those static methods so that I could figure out what was modifying my PromiseToken.

It turns out what happens is I have an error handler function that is triggered in the event of an HTTP error. This sends the config params to the logger, which adds an entry to the DB. The config params include a reference to my PromiseToken, which means that `sql.json()`, calls `deepFreeze` on the token.

Thus, attempting to modify its state later is prevented.

## Expected Behavior

The docs describe `sql.json()` as: `Serializes value and binds it as a JSON string literal`. The general expectation is that a serializer is a pure function and that it will not mutate the object passed to it. 

As an example, imagine if `JSON.stringify()` mutated the object passed to it.

## Possible Solutions

1. **(simple) Add a warning to the documentation that this object will recursively mutate the object passed to it.**
   This is a good thing to do to start, but I hope that I can be persuasive in advising against having a serializer mutate the params.

2. **Create a deep copy of `value` before calling deepFreeze.**
   This is a viable solution, however, there is a cost to deep copy that is unnecessary considering that this can be performed synchronously (see solution 3)

3. **Synchronously serialize the information immediately when `sql.json` is called**
   This is bound to be the most performant method and will not require freezing the object, as it is performed synchronously.

## Steps to Reproduce

Here's a contrived example showing the problem:

```ts
// Get our record info
const rec = await db.one<MyDbRecord>(sql`SELECT * FROM mytable`);

// Add a log
await db.query(`INSERT INTO logs (record) VALUES(${sql.json(rec)})`);

// Update our record
rec.enabled = false; // Error
// ... logic to UPDATE mytable with rec would go here, but it won't get this far
```

## Logs

```
[DEBUG]:  Ran Spied on Function: Object.freeze()
          PromiseToken {
            cancelled: false,
            children: Set(0) {},
            _resolveWait: undefined,
            onCancel: undefined,
            waitParent: undefined,
            waitPromise: undefined
          }
          at Function.Object.<computed> (/home/ron/node/packages/core/dist/nld/bootstrap.js:83:35)
          at deepFreeze (/home/ron/node/node_modules/slonik/dist/utilities/deepFreeze.js:24:10)
          at deepFreeze (/home/ron/node/node_modules/slonik/dist/utilities/deepFreeze.js:28:7)
          at deepFreeze (/home/ron/node/node_modules/slonik/dist/utilities/deepFreeze.js:28:7)
          at deepFreeze (/home/ron/node/node_modules/slonik/dist/utilities/deepFreeze.js:28:7)
          at deepFreeze (/home/ron/node/node_modules/slonik/dist/utilities/deepFreeze.js:28:7)
          at deepFreeze (/home/ron/node/node_modules/slonik/dist/utilities/deepFreeze.js:28:7)
          at Function.sql.json (/home/ron/node/node_modules/slonik/dist/factories/createSqlTag.js:98:38)
          at /home/ron/node/packages/core/dist/components/logger/logger.js:138:90
          at Logger.logWorker (/home/ron/node/packages/core/dist/components/logger/logger.js:152:15)
          at /home/ron/node/packages/core/dist/components/logger/logger.js:68:64
          at APIClient.errorHandler (/home/ron/node/packages/core/dist/components/api/api-client.js:137:50)
          at runMicrotasks (<anonymous>)
          at processTicksAndRejections (internal/process/task_queues.js:93:5)
          at async AsyncTask.action (/home/ron/node/packages/core/dist/components/api/api-client.js:104:17)
```

nonara:
Hi Gajus!

To address the major issue, I added a token check to `isSubjectFreezable`.

In the course of this PR, one thing came up that needs attention.

There is a function `isSqlToken` which formerly implemented an improper type-guard return type. Its return type was `subject is typeof Tokens[number]` which evaluates to a union of string literals (the token literals), while the actual passed parameter is logically required to be an object.

In short, it was checking an object with type `SqlTokenType`, yet improperly telling TS it was validating to a string literal. I see that `@ts-expect-error` was used to compensate for that in `createSqlTag`

To address that I took the following actions:

1. Created a new function called `isSqlTokenType`, which checks object logic (same as former `isSqlToken`)
2. Changed `isSqlToken` logic to check the passed variable matches any of the token string literals.

## Impact

This helps correct a type issue in the code and the naming convention more accurately depicts the typechecker function. However, I have noted that `isSqlToken` is exported from the main package (although undocumented).

That in mind, the impact potential is that anyone who may be using the `isSqlToken` helper from the package would need to switch to the semantically corrected `isSqlTokenType`.

Being that this is undocumented, I'm not sure what weight this holds with you, so I wanted to give you the full background.

## Alternative

If you decide you do not want to incur that risk, I can remove the new function and simply update the typechecker return value on the original `isSqlToken` to the correct type. This would preserve the existing functionality.

If you'd like it or anything else altered, let me know, and I'd be happy to make the adjustments!


nonara:
Associated issue #227

gajus (commented):
Can you think of a good reason for keeping `deepFreeze` to begin with?

nonara:
I can see value to having immutability for property containing actual SQL as a protective measure to make sure it's not easy to circumvent the protection afforded by this library. 

Beyond that I can't think of a reason for locking anything. Though, I'm not sure what the original reasoning was, so I may be missing something.

Locking the sql property can be accomplished without deepFreeze and non-recursively by simply configuring the property descriptor to `enumerable: false`

ie.
```ts
const g = { a: 'hello' };
Object.defineProperty(g, 'a', { configurable: false, writable: false });

g.a += '1' // TypeError: Cannot assign to read only property 'a' of object '#<Object>'
```

Would you like to see it with deepFreeze removed?

gajus:
The original reason for deep freeze was because we used Symbols to validate that SQL tag was generated using Slonik method. There was literally no other way to add a token other than using `sql.*` methods. However, it soon became clear that we need to support https://www.npmjs.com/package/slonik-sql-tag-raw, which threw our the benefit of that protection out the window (because it can be circumvented using sql raw).

As it stands, I don't see any benefit to using `deepFreeze`.

If you have capacity to make the change, I would suggest to just entirely remove `deepFreeze`. Otherwise, I can attend this in the next couple of days.

nonara:
@gajus Sounds good. I'll take a look in a bit.

How would you like to handle the `isSqlToken` issue? Options on the table are:

1. Revert to previous implementation but fix the type signature
2. Revert and keep broken signature
3. Keep the semantically corrected changes

While I'm at it, I wrap slonik and create a `maybeMany` method which I end up using a _lot_. I noticed that in my code, I kept having to write a try-catch for the sole purpose of checking `NotFoundError` with `many`. 

If you're okay with it, I'd also be glad to add `maybeMany`, which returns undefined if no records found. I can then rebase to two commits for merging to master (one `fix` and one `feat`)

gajus:
> While I'm at it, I wrap slonik and create a maybeMany method which I end up using a lot. I noticed that in my code, I kept having to write a try-catch for the sole purpose of checking NotFoundError with many.

Just use `any` method.

gajus:
> Revert to previous implementation but fix the type signature

This sounds reasonable.

I am not sure I understand what is the downside?

nonara:
> Just use any method.

That would definitely work for most folks. In our wrapper, we overload slonik's methods to allow passing a class which acts as a model. It has some metadata that allows a query to be constructed. Minimal abstraction, but it's useful. So because the type is inferred from the model, the `any` method doesn't apply to my use-case, but since it does for everyone else, I agree that there is no reason to add it to the main library.

>> Revert to previous implementation but fix the type signature

> This sounds reasonable.

Will do. 

> I am not sure I understand what is the downside?

It's more of a matter of semantics. No downside in terms of functionality. In TypeScript, it's common for a type guard's name to represent the name of the type you're checking. In this case you have `SqlToken`, which is a union of string literals, and `SqlTokenType`, which is an object with a property that is `SqlToken`. One would expect that `isSqlToken` would be validating a string against the union of literals as opposed to an object.

src/factories/createSqlTag.ts
```diff
@@ -91,75 +90,81 @@ export const createSqlTag = <T = QueryResultRowType>() => {
...
+      writable: false,
     });
 
     return query;
   };
 
   sql.array = (
     values: ReadonlyArray<PrimitiveValueExpressionType>,
-    memberType: string | SqlTokenType,
+    memberType: TypeNameIdentifierType | string | SqlTokenType,
```
nonara, line 110: Adding `TypeNameIdentifierType` here should allow auto-complete (intellisense) to supply possible options

gajus:
Thanks! Looks good

Issue #252, opened 2021-02-08
Transaction binder does not allow streams


issa-tseng:
`createPool(…)` results in an object that has `stream()` on it, but `pool.transaction(…)` does not.
I am happy to create a PR plumbing that through, but I'm wondering first if there is some explicit reason the feature is not made available through that interface.

gajus:
I cannot recall a specific technical reason why streams would not be available for transactions. If I had to guess, it was by design.

Transactions are supposed to be as atomic as possible. Keeping transactions open for a long time is a bad practice, e.g. it will block all vacuum related operations.

gajus:
I've closed this, but if you have a use case that you want to discuss, I am all ears.

issa-tseng:
for better or for worse, we typically run each of our tests in a transaction that we roll back. this cuts our test time from ~13m to ~38s. the lack of the interface is the only thing preventing me from finishing this transition from knex.

(in general i understand the desire for safety and good principles but in practice sometimes not having these interfaces just means people have to do even nastier and more evil things to work around them. i've had to add a lot of handling i don't feel so sure about to make date values work, for instance, especially since i'm migrating an existing project that uses such values.)

gajus:
That's understandable.

If you PR, I am happy to do add it.

PR #254, opened 2021-02-10
feat: allow .stream() on transaction binders (#252)

issa-tseng:
as discussed in #252. happy to make any other changes of course. and also happy to not include the test if it's frivolous.


issa-tseng:
not sure why my typescript wants different things than travis, but oh well. sorry it took me a minute to get the build fixed on this pr. it's good to go now.

PR #260, opened 2021-03-08
tuple to be locked was already moved to another partition due to concurrent update error handling

skarbovskiy:


test/slonik/connectionMethods/query/query.ts
```diff
@@ -191,6 +192,16 @@ test('57P01 error causes the connection to be rejected (IMPLICIT_QUERY connectio
   t.true(error instanceof BackendTerminatedError);
 });
 
+test('tuple to be locked was already moved to another partition due to concurrent update error handling', async (t) => {
+  const pool = createPool();
+
+  pool.querySpy.rejects(new Error('tuple to be locked was already moved to another partition due to concurrent update'));
```
gajus, line 198: Is there a way to trigger this error in the integration tests?

Triggering it like this gives no guarantees that this will work with different Postgres versions.

skarbovskiy: I tried to simulate race condition based on these tests https://github.com/postgres/postgres/commit/f16241bef7cc271bff60e23de2f827a10e50dde8#diff-0db9ceae2f40743b6e27eb1589e659b7c4a66fdef742d9b3d5fcf543ede9850cR71 but failed. As far as I see this can only be achieved by manualy setting right order of update and commit calls inside postgres core.

gajus: Create two connections and and make sure to wait for the completion of each step to force the situation.

* Create conn A
* Create conn B
* A:BEGIN
* B:BEGIN
* A:UPDATE...
* B:UPDATE (don't wait for this)
* A:COMMIT
* B:UPDATE

should be rejected with the error.

skarbovskiy: thanks

PR #262, opened 2021-03-19
 added delay in concurrent partitioned table updates test to ensure query is processed before concurrent commit is called

skarbovskiy:


test/slonik/integration.ts
```diff
@@ -1094,6 +1094,7 @@ test('Tuple moved to another partition due to concurrent update error handled',
         );
         t.true(error instanceof TupleMovedToAnotherPartitionError);
       });
+      await delay(1000);
```
gajus, line 1097:
```suggestion
      // Ensures that query is processed before concurrent commit is called.
      await delay(1000);
```

gajus:
Thank you

skarbovskiy:
can u release this please?)

PR #263, opened 2021-03-25
feat(types): remove readonly from result types

### LINKED ISSUE #218: Question: Why are query results marked as `readonly`?
(sorry this isn't a bug, but can't figure out how to remove the label)

I just upgraded to version `23.0.1`. I use typescript and was excited to see that the library had been re-written in typescript.
Overall the upgrade seemed to go smoothly. However, on my `.many<MyObjectType>()` calls, I had type errors saying that the result was marked as readonly, and could not be cast to mutable type `MyObjectType[]`. 

I'm writing just to ask why this is the case. I'm able to get around this and return my mutable array type by spreading the results into a new array. 

Here's a more complete example: 

```typescript
// Original code, with type error
async function getResponses(): Promise<UserAssessmentResponse[]> {
    const responses: UserAssessmentResponse[] = await this.db.pool.many<UserAssessmentResponse>(sql`
        SELECT *
        FROM user_assessment_response
        WHERE user_assessment_id = ${userAssessmentId};
        `);

    /** TYPE ERROR: 
    The type 'readonly UserAssessmentResponse[]' is 'readonly' and cannot be assigned to the mutable type 
    'UserAssessmentResponse[]
    */
    return resposnes
}

```

```typescript
// "fixed" code
async function getResponses(): Promise<UserAssessmentResponse[]> {
    const responses: readonly UserAssessmentResponse[] = await this.db.pool.many<UserAssessmentResponse>(sql`
        SELECT *
        FROM user_assessment_response
        WHERE user_assessment_id = ${userAssessmentId};
        `);

    return [...responses]
}

```

mmkal:
Fixes #218 

Summarising some of the arguments here. As a simple example of the motivation - it allows passing slonik results to regular utility functions:

```ts
const first3 = <T>(list: T[]) => list.slice(0, 3)

const myResults = await pool.any(sql`select * from sometable`)

console.log(first3(myResults))
```

Currently, the above errors unless `first3` is changed to `<T>(list: readonly T[]) => ...`, or the results array is sliced, or uses a type assertion. All of these are unnecessary overhead and will likely lead to developers taking a shortcut and using `any`.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,7 +64,7 @@ export type QueryResultType<T> = {
   readonly fields: readonly FieldType[],
   readonly notices: readonly NoticeType[],
   readonly rowCount: number,
-  readonly rows: readonly T[],
+  readonly rows: T[],
 };
 
 export type InternalDatabasePoolType = any;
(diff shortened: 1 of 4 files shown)


mmkal:
Closing in favour of #266

PR #265, opened 2021-03-29
Improve type safety for query methods

dwickern:
The multi-column variants (`any`, `many`, `maybeOne`, `one`, `query`) always return a hash.
The single-column variants (`anyFirst`, `manyFirst`, `maybeOneFirst`, `oneFirst`) always return a single column type.

This PR makes it harder to use the wrong one.

```ts
// correct
connection.any<{ col1: number, col2: string}>(sql`...`)
connection.anyFirst<number>(sql`...`)

// incorrect
connection.any<string>(sql`...`)
connection.anyFirst<{ foo: number }>(sql`...`)
```

This change caught one such bug in the test suite.

### DIFF
--- a/src/factories/createSqlTag.ts
+++ b/src/factories/createSqlTag.ts
@@ -162,6 +162,6 @@ sql.unnest = (
   };
 };
 
-export const createSqlTag = <T = QueryResultRowType>(): SqlTaggedTemplateType<T> => {
+export const createSqlTag = <T extends QueryResultRowType = QueryResultRowType>(): SqlTaggedTemplateType<T> => {
   return sql;
 };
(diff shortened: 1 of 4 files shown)


xiaoyu-tamu:
after upgrading to v23.6.4, query helper(pool.one, pool.many) cannot infer type from sql tag.

23.6.4
image


23.6.3
image

felamaslen:
Related issue: https://github.com/gajus/slonik/issues/275

The issue here is bigger than just `json`. It's possible to add any custom type parsers, e.g. you could parse `int8` into a `Uint8Array` if you really wanted to. So surely the query result should support any arbitrary type in the generic.

PR #267, opened 2021-04-03
Allow json types in query methods

alitaheri:
In #265 some type safety were enforced. But it caused json types to be considered invalid.

```js
connection.any<{ col1: number, col2: {myJsonColumn: string}}>(sql`...`)
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -273,7 +273,7 @@ export type UnnestSqlTokenType = {
 };
 
 export type PrimitiveValueExpressionType =
-  Buffer | boolean | number | string | readonly PrimitiveValueExpressionType[] | null;
+  Buffer | boolean | number | string | object | readonly PrimitiveValueExpressionType[] | null;
 
 export type SqlTokenType =
   | ArraySqlTokenType
--- a/src/types.ts
+++ b/src/types.ts
@@ -273,7 +273,7 @@ export type UnnestSqlTokenType = {
 };
 
 export type PrimitiveValueExpressionType =
-  Buffer | boolean | number | string | object | readonly PrimitiveValueExpressionType[] | null;
+  Buffer | boolean | number | object | string | readonly PrimitiveValueExpressionType[] | null;
 
 export type SqlTokenType =
   | ArraySqlTokenType


dwickern:
Good catch. It looks like `json` and `jsonb` are supported by [node-postgres](https://node-postgres.com/features/types), which explains the lack of code+tests for it in slonik. According to that page, we might also be missing `Date` as a column type.

The `object` type means "any non-primitive", which includes things like functions and classes. I think this is the best way we could represent a JSON column:
```ts
type JsonValue = string | number | boolean | null | { [prop: string]: JsonValue } | JsonValue[];

export type PrimitiveValueExpressionType =
  Buffer | boolean | number | JsonValue | string | readonly PrimitiveValueExpressionType[] | null;
```

aengl:
@dwickern That does indeed look like a better alternative to `object`. I updated my local version with your proposed changes and it addresses the errors introduced by #265 in my codebase.

gajus:
The only reason this isn't merged is because of the failing linting.

gajus:
@alitaheri Do you plan to fix the PR?

felamaslen:
Related issue: https://github.com/gajus/slonik/issues/275

The issue here is bigger than just `json`. It's possible to add any custom type parsers, e.g. you could parse `int8` into a `Uint8Array` if you really wanted to. So surely the query result should support any arbitrary type in the generic.

CSNWEB:
I am not sure arbitrary types should be allowed there. It happens quite often, to me at least, that I put in an object type by accident. The current type prevents developers from writing an object types where there most probably shouldn't be one. Moreover the schema of the json is probably not checked by the database, so a different shape might be returned. In any case I think this can be moved forward so I tried to fix the outstanding issues in my own fork in #276

dwickern:
We could fix this issue and #275 by adding a registry for custom types:

```ts
export interface SqlTypeRegistry {
  _if_empty: never;
}
type CustomSqlType = SqlTypeRegistry[keyof SqlTypeRegistry];

export type PrimitiveValueExpressionType =
  Buffer | CustomSqlType | boolean | number | string | readonly PrimitiveValueExpressionType[] | null;
```

The user adds their custom types via declaration merging:

```ts
type Json = string | number | boolean | null | { [prop: string]: Json } | Json[];

declare module 'slonik' {
  interface SqlTypeRegistry {
    // enable json columns
    json: Json;

    // enable types for custom parsers
    date: Date;
    int8: Uint8Array
  }
}
```

And then `Json`, `Date` and `Uint8Array` would be recognized as column types. I'm leaving JSON as opt-in here since you lose a lot of type safety enabling it.

WDYT?

mmkal:
IMO arbitrary types should be allowed. If you're hand-writing the generic, it might be wrong anyway:

```sql
let x = await pool.one(sql<{ bar: string }>`
  select 123 as foo
`)
```

`x` will be typed as `{ bar: string }` when it should be `{ foo: string }`. I'm working on a library that will codegen the correct types by running `\gdesc` on queries at build time to take away human error possibility here, but I don't really see how adding arbitrary json to the query return signature increases or decreases the risk. The SqlTypeRegistry idea is neat, but it's really just a patch of slonik's types not quite matching runtime/reality.

Also note - the TypeScript team made the decision quite long ago that it isn't worth having a fancy `json` type because it's so close to `any` anyway: https://github.com/microsoft/TypeScript/issues/1897#issuecomment-72546389

If we did want to get really fancy, we could make type parsers generic, then each `DatabasePoolType` could collect all the return types of the type parsers and allow queries run by that pool/sub-connections/-transactions to return columns of those types. But it'd be a lot of complexity, I don't think it's worth it right now, the return type should just be unrestricted.

felamaslen:
> I'm working on a library that will codegen the correct types by running `\gdesc` on queries at build time to take away human error possibility here

This would be fantastic. Have you seen [pgtyped](https://github.com/adelsz/pgtyped)? It seems to be doing something similar.

felamaslen:
>  This issue has been resolved in version 23.7.0 
> 
> The release is available on:
> 
> * [npm package (@latest dist-tag)](https://www.npmjs.com/package/slonik/v/23.7.0)
> * [GitHub release](https://github.com/gajus/slonik/releases/tag/v23.7.0)
> 
> Your **[semantic-release](https://github.com/semantic-release/semantic-release)** bot

This release does not fix the issue.

On [line 376 of `src/types.ts`](https://github.com/gajus/slonik/blob/v23.7.0/src/types.ts#L376), the generic of `QueryFunctionType` still extends `ExternalQueryResultRowType`, which is `Record<string, QueryResultRowColumnType>`.

`QueryResultRowColumnType` is `PrimitiveValueExpressionType` which does not allow arbitrary types.

mmkal:
> > I'm working on a library that will codegen the correct types by running `\gdesc` on queries at build time to take away human error possibility here
> 
> This would be fantastic. Have you seen [pgtyped](https://github.com/adelsz/pgtyped)? It seems to be doing something similar.

I hadn't, looks interesting. What I'm working on is similar, but specific to slonik. My aim is to have it understand things like type parsers, as far as possible. And also to work for queries within typescript e.g.

```ts
await pool.query(sql`select foo from bar`)
```

adevine:
Hey folks, I originally left a comment on another issue but moving here as this is where most of the discussion is. Like @felamaslen commented, I also had the same issue that 23.7.0 did not fix all the breaking changes introduced by https://github.com/gajus/slonik/pull/265. My suggestion is as follows:

1. Can we just fully revert 265 for the time being? Given all the issues that popped up after it broke code for a lot of people.
2. In general, I think one should be very careful whenever narrowing types based on return value from queries. As stated in this thread, given JSON fields and custom type parsers you can basically return anything from a SQL query, so narrowing types is almost by definition going to be backwards incompatible, and as such should require a major version change.
3. I agree with @mmkal's comment above - arbitrary types should be allowed in the slonik typedefs. I don't think I should have to 'patch' a library with a custom type definition if I'm using something like JSON types that is fully supported by postgres. I think the idea of a codegen tool to run at buildtime is very cool, but again that leaves me with the freedom to decide if I want to use it.

mmkal:
@felamaslen @gajus @adevine I published [@slonik/typegen](https://www.npmjs.com/package/@slonik/typegen) which now runs at build time. It's a little rough around the edges, but quite low risk to try out since it's just a CLI which adds types to your existing project. There are no associated production dependencies. It'd be great to get any feedback on it for a real-world project. The output is somewhat noisy, and it might fail to type certain queries, but in general fails gracefully and just leaves them as-is.

As a follow-up I'd like it to also help enforce parameter types, i.e.:

```ts
sql<Foo>`
  select *
  from foo
  where a = ${x} and b = ${y}
`
```

It would be possible for the library to figure out that `x` should be a `string` and `y` a number, based on columns `a` and `b`. It would require a slight tweak to slonik's types, though, so I will raise a PR.

___

In the meantime, I think the spirit of this PR is still valid. @gajus what I learned in building the typegen library: in my opinion, all generics should be on the `sql` tag, and the various query methods should just inherit behaviour from there. Any decisions about restrictions on `sql`'s typeargs don't need to be duplicated in the various query methods. They just need to be generic for type inference to work properly. And the typescript compiler struggles to infer properly when the `extends` and default typeargs are added to some of the methods.

All of this is quite hard to talk about in the abstract, so here's a PR which reproduces the issue directly with some tests, you can pull the branch to see the problem in your IDE: https://github.com/gajus/slonik/pull/282. The CI job should also catch the problem. The first commit 5b17a71 reproduces the problem and the second commit 5e37def fixes it.

adevine:
Thanks very much @mmkal and @gajus!

Misha, can't wait to try out your typegen project. IMO the general design of this is exactly how I'd want something like this to work (though I still need to read more of your doc to understand the config details). I'll let you know if we find anything when we try it out.

PR #276, opened 2021-05-12
Fix #267 Allow json types in query methods

CSNWEB:
As there was no response in #267 by @alitaheri, I tried to fix the ts errors myself here.


CSNWEB:
I believe the failing test is also failing on master at the moment.

Issue #270, opened 2021-04-09
[TypeScript] Cannot use interface type because of ExternalQueryResultRowType constraints


davidskuza:
If we want to query using simple plain interface TS compiler complains about type constraints: 

Type 'TestInterface' does not satisfy the constraint '**ExternalQueryResultRowType**'.

## Expected Behavior
To work

## Current Behavior
Doesn't work

## Steps to Reproduce
[...]

Issue #268, opened 2021-04-08
Date typeParser not working any more


bencehornak:
Let's say I have a type parser that outputs a `Date` object, and I have a `timestamptz` field in the db that is mapped with it. I want to use a TypeScript interface for the returned object.

```sql
const pool = createPool('postgres://', {
  typeParsers: [
    createTimestampTypeParser()
  ]
});
interface Foo {
  bar: Date;
}
await pool.one<Foo>(sql`SELECT NOW()::timestamp AS bar`);
```

## Expected Behavior
It is expected to compile successfully.

## Current Behavior
I get a compile error since `slonik@23.6.4`:
```
Type 'Foo' does not satisfy the constraint 'ExternalQueryResultRowType'.
  Property 'bar' is incompatible with index signature.
    Type 'Date' is not assignable to type 'PrimitiveValueExpressionType'.
      Type 'Date' is missing the following properties from type 'Buffer': write, equals, compare, copy, and 77 more.ts(2344)
```

## Possible Solution
A new constraint has been added in this version to the generic type parameter of the `QueryOneFunctionType` (and to all its friends):
```diff
--- src/types.ts    v23.6.3
+++ src/types.ts    v23.6.4
@@ -1,4 +1,4 @@
-export type QueryOneFunctionType = <T = ExternalQueryResultRowType>(
+export type QueryOneFunctionType = <T extends ExternalQueryResultRowType = ExternalQueryResultRowType>(
        sql: TaggedTemplateLiteralInvocationType<T>,
        values?: PrimitiveValueExpressionType[],
   ) => Promise<T>;
```

`Foo` is not compatible with `ExternalQueryResultRowType` due to the `Date` field, as the compiler error suggests.

The solution could be to revert this `extends` constraint.

## Steps to Reproduce
**Update:** I created a test case based on both [`v23.6.3`](https://github.com/bencehornak/slonik/tree/test-typing-v23.6.3) and [`v23.6.4`](https://github.com/bencehornak/slonik/tree/test-typing-v23.6.4). The first one compiles, the second one doesn't.

![Screen Shot 2021-04-09 at 11 11 35](https://user-images.githubusercontent.com/7274955/114159066-aec64080-9925-11eb-9e18-11f0bd6581df.png)

bencehornak:
Fixed in v23.7.0 🎉 
@gajus Many thanks!

PR #282, opened 2021-06-25
Fix types for rows with jsonb fields

mmkal:
- [x] test cases reproducing the issue identified in #267 and a few other issues.
- [x] a fix in the types
- [x] full listing of issues and PRs this can close

The first commit 5b17a71 reproduces the problem and the second commit 5e37def fixes it. **I would recommend a squash-and-merge commit to avoid having the intermediate "red" commit in master**.

Since no existing type-level tests regressed, we can be pretty confident that this won't break any further use cases (e.g. we can see that the type restrictions actually _weren't_ helping make the overall type-safety any better, since we have plenty of tests that ensure the various query methods return appropriate types). If anyone can think of more type-level test cases to add, please feel free to comment.

Closes #267 
Closes #276
Related: #268
Related: #270
Related: https://github.com/gajus/slonik/commit/09c7db1611214dfa44d385b84dc7b034e745124b
Related #265 (the PR that caused all the chaos!)

As a follow-up, I think there should be something added to the docs that pushes users away from specifying generics on the query helper methods, preferring instead to put them on the `sql` tag itself, since that's what actually determines the resultant SQL query and the type associated with it, and it allows all queries to be typed in a consistent way. Each of `.query`, `.any`, `.anyFirst`, `.one`, `.oneFirst` have their own unique generic formats, which is very hard - and unnecessary - to keep track of (I say that as the person who originally wrote most of them!). Something along the lines of:

```ts
// bad
await pool.query<{ foo: string }>(sql`select foo from bar`)

// good
await pool.query(sql<{ foo: string }>`select foo from bar`)
```

👆 the `// good` pattern is also how [@slonik/typegen](https://npmjs.com/package/@slonik/typegen) now works.

Side note: @gajus I'm happy to help review future PRs to do with typescript/anything else if it'd be useful!

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -356,16 +356,14 @@ export type InternalNestedTransactionFunctionType = <T>(
   transactionDepth: number,
 ) => Promise<T>;
 
-type ExternalQueryResultRowType = Record<string, QueryResultRowColumnType>;
-
 // eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/consistent-type-definitions, @typescript-eslint/no-unused-vars
 export interface TaggedTemplateLiteralInvocationType<Result extends UserQueryResultRowType = QueryResultRowType> extends SqlSqlTokenType { }
 
-export type QueryAnyFirstFunctionType = <T extends QueryResultRowColumnType = QueryResultRowColumnType, Row extends Record<string, T> = Record<string, T>>(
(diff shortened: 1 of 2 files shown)


gajus:
Thank you

> As a follow-up, I think there should be something added to the docs that pushes users away from specifying generics on the query helper methods, preferring instead to put them on the sql tag itself, since that's what actually determines the resultant SQL query and the type associated with it, and it allows all queries to be typed in a consistent way.

This makes sense.

PR #299, opened 2021-09-30
bigint parser now returns only BigInt.

muratgozel:
Hi,

Instead of conditionally returning BigInt (as you mention in the todo comment), the bigint type should always return BigInt so developers always know how to deal with it.

BigInt support in node.js exists as of version [10.8.0](https://node.green/#ES2020-features-BigInt).

I haven't reviewed the codebase much. Would like to know your opinion and hope this won't break anything.


gajus:
This is a lot more involved change than this, so closing this PR.

Please continue discussing in one of the open issues https://github.com/gajus/slonik/issues?q=is%3Aissue+is%3Aopen+bigint

Issue #239, opened 2020-12-04
Timestamp type parser breaks if timezone is not GMT


NicolasCaous:
I'm from São Paulo, so my default timezone is UTC-3. Whenever I try to use the `timestamp` field, its values are returned ahead of time (exactly the time difference between my local time and GMT).

## Expected Behavior
The timestamp returned from a `select` query should be the time stored inside the database.

## Current Behavior
The timestamp returned from a `select` query is ahead of the true value stored in the database by exactly the difference from my local time to GMT.

## Possible Solution

Timezone conversion error, can't pinpoint because I couldn't understand the library source code. However, `timestamptz` works flawlessly. So a possible solution is to check why `timestamptz` behavior is different from `timestamp` when the GMT timezone is used.

## Steps to Reproduce
1 - Be in a system with a different timezone from GMT

2 -
```javascript
// https://momentjs.com/timezone/
const moment = require("moment-timezone");

await pool.query(sql`create table "test" ( "field" timestamp )`)

let date = new Date()
await pool.query(sql`insert into "test" ("field") values (${moment.utc(date).format("YYYY-MM-DD HH:mm:ss.SSS")})`)

let dbDate = new Date((await pool.query(sql`select * from "test"`)).rows[0].field)

console.log(date)
console.log(dbDate)
```
## Logs
```javascript
Welcome to Node.js v14.15.1.      
Type ".help" for more information.
> const moment = require("moment-timezone");
undefined
> await pool.query(sql`create table "test" ( "field" timestamp )`)
{ command: 'CREATE', fields: [], notices: [], rowCount: 0, rows: [] }
> let date = new Date()
undefined
> await pool.query(sql`insert into "test" ("field") values (${moment.utc(date).format("YYYY-MM-DD HH:mm:ss.SSS")})`)
{ command: 'INSERT', fields: [], notices: [], rowCount: 1, rows: [] }
> let dbDate = new Date((await pool.query(sql`select * from "test"`)).rows[0].field)
undefined
> console.log(date)
2020-12-04T22:26:14.113Z
undefined
> console.log(dbDate)
2020-12-05T01:26:14.113Z
undefined
```

![Sem título](https://user-images.githubusercontent.com/24411365/101221737-12ceb480-3667-11eb-89d6-31f561185792.png)

My local time: 2020-12-04 19:26:00
UTC time: 2020-12-04 22:26:00

aklotos:
[...]

PR #304, opened 2021-10-18
fix: remove UTC from timestamp with time zone

C-l-o-u-d:
`UTC` was wrongly added to the `timestamp with time zone (timestamptz)`

i.e. result expession will be `Date.parse('2004-10-19 10:23:54+04 UTC')`

ref: https://github.com/gajus/slonik/issues/239#issuecomment-946177310


gajus:
Thank you

Issue #300, opened 2021-09-30
Question: How to build a query with a dynamic where clause?


lamontadams:
I'm trying to generate a dynamic where clause something like this: 
```
const field1 = 1
const field2Clause = someValue ? ` and field2=${someValue}` : ''
sql`select * from table where field1=${field1}${field2Clause}`
```
and I'm unable to work out how to do it. As written that results in a syntax error because parameter tokenization results in a query like `select * from table where $1$2`

Looking at the [query-building section of the docs](https://github.com/gajus/slonik#slonik-query-building) I don't see any support mentioned, but there's a suggestion to see the dev.to post [Dynamically generating SQL queries using Node.js](https://dev.to/gajus/dynamically-generating-sql-queries-using-node-js-2c1g) which shows the use of a whole bunch of helper functions (`sql.raw` in the "Injecting dynamic SQL fragments" section for one) that apparently have been removed.

So, is this possible? Do I need to install some other package to have `sql.raw` available?

Suggestion: I don't want to pile on, but the project docs here really do leave something to be desired. If the dev.to post the docs link to is outdated, then the suggestion to read it should be removed. Otherwise update the docs so it's clear how to make use of the information it contains. The current situation is just confusing and irritating.

albertreed:
I believe you want `sql.identifier`.

albertreed:
(sorry for the initial misread of your question).

To accomplish this, I'd just build a query that does the whole thing rather than putting the logic in JS. Not sure if that's best practice or not.

lamontadams:
No worries, sorry for my kinda snarky reply.

Building SQL with elaborate boolean logic to only apply supplied values is a solution, but part of the reason we use query builders is to, you know, build dynamic queries. ;)

mmkal:
I believe this would work:

```ts
const field1 = 1
const field2Clause = someValue ? sql` and field2=${someValue}` : sql``
sql`select * from table where field1=${field1}${field2Clause}`
```

But +1 on the docs update request, I think I discovered that by accident and now just remember it.

gajus:
The original question has been answered my @mmkal .

Regarding the suggestions to improve documentation, please contribute to the documentation.

It is a lot simpler to see what's need to be added to the documentation for someone who is just figuring it out, than for someone who have been using the API for a long time.

I will gladly support anyone who wants to contribute to the docs.

albertreed:
I'll put together a PR to capture this pattern in the docs. 

[...]

PR #305, opened 2021-10-27
Added example for building a dynamic query to documentation.

albertreed:
See: https://github.com/gajus/slonik/issues/300

### DIFF
--- a/.README/RECIPES.md
+++ b/.README/RECIPES.md
@@ -73,5 +73,16 @@ masterPool.query(sql`SELECT 1`);
 
 // This query will use `postgres://master` connection.
 masterPool.query(sql`UPDATE 1`);
+```
 
+### Building a dynamic query
+
+To build a query dynamically, assemble the dynamic portions of the query inside of a `sql` tag. You can then combine those clauses inside of other `sql`tags to be used to execute queries.
+
+For example:
+
+```js
+const field1 = 1;
+const field2Clause = someValue ? sql` and field2=${someValue}` : sql``;
+const finalQuery = sql`SELECT * FROM table WHERE field1=${field1}${field2Clause}`;
 ```


albertreed:
@gajus Any thoughts here? Was going to take a small handful of documentation items from past questions but wanted to be sure I'm on the right track before going wider.

gajus:
Frankly, I found this addition confusing.

If it was me, I would have written something along the lines of:

### Building dynamic queries

Queries are built using `sql` tag and those tags can be nested to create dynamic queries, e.g.

```ts
const whereFragment = sql`WHERE bar='baz'`;
const fooQuery = sql`SELECT * FROM foo ${whereFragment}`;
```

In the above example `whereFragment` is embedded in `fooQuery`. Therefore, the resulting query is equivalent to:

```sql
SELECT *
FROM foo
WHERE bar='baz'
```

AndrewJo:
In addition, I find that `sql.join` is better suited for more complex dynamic queries and scales better if you have more than just a couple of clauses in addition to other query building blocks provided by Slonik.

Expanding the proposal a bit to add additional clauses:

```js
const field1 = 1;
const field2Clause = someValue ? sql` and field2=${someValue}` : sql``;
const field3Clause = otherValue ? sql` and field3=${otherValue}` : sql``;
const finalQuery = sql`SELECT * FROM table WHERE field1=${field1}${field2Clause}${field3Clause}`;
```

If you have more than a couple of fields, `sql.join` especially makes the code cleaner compared to tacking on `${}` one after another:

```js
const whereClauses = [sql`field1 = ${field1}`];
if (someValue) whereClauses.push(sql`field2 = ${someValue}`);
if (otherValue) whereClauses.push(sql`field3 = ${otherValue}`);
const finalQuery = sql`SELECT * FROM "table" WHERE ${sql.join(whereClauses, sql` AND `)}`;
```

See this convoluted example from what I use at my company for generating dynamic JOIN statements:

```js
/**
 * Gets the building for an electrical panel, a circuit breaker, or at a lower level, a device
 * Building has many electrical panels, panels have many circuit breakers.
 */
async function getBuilding(connection, entity, entityId) {
  const joinClauses = [];

  // Switch fallthrough is very useful for constructing joins
  switch (entity) {
    case "devices":
      joinClauses.unshift(sql`INNER JOIN devices ON devices.circuit_breaker_id = circuit_breakers.id`);
    case "circuit_breakers":
      joinClauses.unshift(sql`INNER JOIN circuit_breakers ON circuit_breakers.panel_id = panels.id`);
    case "panels":
      joinClauses.unshift(sql`INNER JOIN panels ON panels.building_id = buildings.id`);
    break;
  }

  return connection.one(sql`
    SELECT * FROM buildings 
    ${sql.join(joinClauses, sql`\n`)
    WHERE ${sql.identifier([entity, "id"]) = ${entityId}`);
}
```

The above function dynamically generates queries based on the input:
```js
getBuilding(connection, "device", 1);

/* Generated query:
SELECT * FROM buildings
INNER JOIN panels ON panels.building_id = buildings.id
INNER JOIN circuit_breakers ON circuit_breakers.panel_id = panels.id
INNER JOIN devices ON devices.circuit_breaker_id = circuit_breakers.id
WHERE "devices"."id" = 1
*/

getBuilding(connection, "panels", 8);

/* Generated query:
SELECT * FROM buildings
INNER JOIN panels ON panels.building_id = buildings.id
WHERE "panels"."id" = 8
*/
```

albertreed:
Hey all - Just wanted to see that I hadn't forgotten this. Digging out of some work stuff but will still get to it (and some other doc requests).

bootrino:
I've spent nearly an hour trying to find out the right way to do this.

It would be enormously helpful and time saving if this bit of documentation was on the README/front page/ right up the front of the documentation.  This is precisely the information I needed

https://github.com/gajus/slonik/pull/305#issuecomment-968994299

Even better if it continued on to provide this information:

https://github.com/gajus/slonik/pull/305#issuecomment-973324847

This thread should be the star paragraph of the documentation, not stuck in an issue thread.

gajus:
@bootrino Please consider contributing documentation improvements.

For those who have been using this library for 5 years+ everything becomes obvious, so it is really hard to think through the eyes of a first time user.

gajus:
Wrote an article walking through query building using Slonik.

https://contra.com/p/AqZWWoUB-writing-composable-sql-using-java-script

PR #308, opened 2021-11-06
Change query retry handling

### LINKED ISSUE #163: Serializable transactions not working as intended
I've created a repo that reproduces this issue: https://github.com/derekchiang/postgres-bug

Basically, I'm running some code in a serializable transaction like this (pseudocode):

```
BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE

const userCount = select count(*) from missions
if (userCount == 0) {
  insert into users default values
}

COMMIT
```

Now, when you run this transaction many times in parallel, you'd think that only one row would be created.  However, as you can see from the test in the repo, more than 1 rows wound up getting created.

It's not clear to me whether this is a bug in Slonik or Postgres, or if I'm fundamentally misunderstanding something.  But I thought I'd put it here in case it's an issue in Slonik.

### LINKED ISSUE #176: Retry failing for implicit pool queries
When using `pool.query` and Postgres throws an error, for example `deadlock detected`, code `40P01` slink attempts a retry. It is possible to have a deadlock even when executing a normal update query, not just a transaction.

## Expected Behavior
We should retry the implicit query.

## Current Behavior
The query mechanism seems to assume there are transactionQueries and we end up with `transactionQueries is not iterable` error. (In node_modules/slonik/dist/routines/executeQuery.js:32:38)
## Steps to Reproduce
Execute several update queries that would deadlock, using implicit `pool.query`

## Logs

```
    TypeError: transactionQueries is not iterable
        at retryTransaction (/Users/ivankerin/Projects/boost-statements-loss/node_modules/slonik/dist/routines/executeQuery.js:32:38)
        at process._tickCallback (internal/process/next_tick.js:68:7) }
```

If I set `transactionRetryLimit: 0`, I get:

```
   error:
    { error: deadlock detected
        at Connection.parseE (/Users/ivankerin/Projects/boost-statements-loss/node_modules/pg/lib/connection.js:614:13)
        at Connection.parseMessage (/Users/ivankerin/Projects/boost-statements-loss/node_modules/pg/lib/connection.js:413:19)
        at Socket.<anonymous> (/Users/ivankerin/Projects/boost-statements-loss/node_modules/pg/lib/connection.js:129:22)
        at Socket.emit (events.js:198:13)
        at Socket.EventEmitter.emit (domain.js:448:20)
        at addChunk (_stream_readable.js:287:12)
        at readableAddChunk (_stream_readable.js:268:11)
        at Socket.Readable.push (_stream_readable.js:223:10)
        at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
      name: 'error',
      length: 323,
      severity: 'ERROR',
      code: '40P01',
      detail:
       'Process 383 waits for ShareLock on transaction 5191; blocked by process 378.\nProcess 378 waits for ShareLock on transaction 5190; blocked by process 383.',
      hint: 'See server log for query details.',
      position: undefined,
      internalPosition: undefined,
      internalQuery: undefined,
      where: 'while updating tuple (59565,14) in relation "supply"',
      schema: undefined,
      table: undefined,
      column: undefined,
      dataType: undefined,
      constraint: undefined,
      file: 'deadlock.c',
      line: '1146',
      routine: 'DeadLockReport' } }
```

### LINKED ISSUE #196: transactionQueries is not iterable
I got this issue often, when my cron run some tasks.
What this error mean? What can I do?

## Expected Behavior

No error

## Current Behavior

> [2020-08-18T09:25:04.046Z] ERROR (50) (@Cron) (#01EG0E5NHKBBWS6GP9TRX4C88Z:GET:/cron/task): InternalQuery / **transactionQueries is not iterable**
SLONIK:
QUERY:
  sql:    SELECT * FROM process_matchator( 5::INTEGER , $1::UUID , $2::INTEGER) as query;
  type:   SLONIK_TOKEN_SQL
  values:
    - null
    - 7

## Steps to Reproduce

I don't know how to reproduce it..

nponiros:
Move transaction retry handling to the transaction and nestedTransaction
methods and change it so that the handler function is called again
instead of just replaying the queries that are part of the transaction.
Fixes #163.

Add a second parameter to the transaction method to allow users to
define a retry limit per transaction. If a retry limit is given for a
transaction then the global defined retry limit for transactions is
ignored.

Change executeQuery so that it retries individual queries that failed
with a transaction rollback error. Only queries that are not part of a
transaction are retried. The number of times a query is retried is
specified by the global queryRetryLimit configuration. Fixes #176.

The above change also fixes #196 since the transactionQueries array no
longer exists.

### DIFF
--- a/.README/QUERY_METHODS.md
+++ b/.README/QUERY_METHODS.md
@@ -375,6 +375,12 @@ ROLLBACK;
 
 Transactions that are failing with [Transaction Rollback](https://www.postgresql.org/docs/current/errcodes-appendix.html) class errors are automatically retried.
 
-A failing transaction will be rolled back and all queries up to the failing query will be replayed.
+A failing transaction will be rolled back and the callback function passed to the transaction method call will be executed again. Nested transactions are also retried until the retry limit is reached. If the nested transaction keeps failing with a [Transaction Rollback](https://www.postgresql.org/docs/current/errcodes-appendix.html) error, then the parent transaction will be retried until the retry limit is reached.
 
-How many times a transaction is retried is controlled using `transactionRetryLimit` configuration (default: 5).
+How many times a transaction is retried is controlled using `transactionRetryLimit` configuration (default: 5) and the `transactionRetryLimit` parameter of the `transaction` method (default: undefined). If a `transactionRetryLimit` is given to the method call then it is used otherwise the `transactionRetryLimit` configuration is used.
+
+#### Query retrying
+
+A single query (not part of a transaction) failing with a [Transaction Rollback](https://www.postgresql.org/docs/current/errcodes-appendix.html) class error is automatically retried.
+
+How many times it is retried is controlled by using the `queryRetryLimit` configuration (default: 5).
--- a/.README/README.md
+++ b/.README/README.md
@@ -37,6 +37,7 @@ Note: Using this project does not require TypeScript. It is a regular ES6 module
 * [Safe value interpolation](#protecting-against-unsafe-value-interpolation).
 * [Transaction nesting](#transaction-nesting).
 * [Transaction retrying](#transaction-retrying)
+* [Query retrying](#query-retrying)
 * Detailed [logging](#slonik-debugging).
 * [Asynchronous stack trace resolution](#capture-stack-trace).
 * [Middlewares](#slonik-interceptors).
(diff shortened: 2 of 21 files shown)


nponiros:
A few things to note:

The blocking/unblocking behaviour for transactions is now different than before. Now we unblock before retrying a transaction. Before the unblock took place after the retry. I don't fully understand the reason for the blocking/unblocking so I'm not sure if this causes any issues.

With the way I implemented the retry mechanism for a single query and transaction the user will see the transaction rollback error if the query/transaction fails despite the retries. Before the user would get an UnexpectedState error. I assumed that this was a bug. Also before the queries were retried until the limit was reached and the result of the last retry was return. I also assumed that this was a bug and now the retry function returns as soon as a query succeeds.

The way I implemented the transaction retry a nested transaction is retried until it reaches its limit and in case the transaction rollback error persist the parent transaction is retried until its limit. I don't know if this is such a good idea. Perhaps it would be better to not retry nested transactions and let the parent transaction do the retry.

gajus:
@nponiros This is really awesome contribution. Thank you for taking time to put it together. I am releasing it as a breaking change.

PR #311, opened 2021-11-18
Add Buffer support to primitive value expressions

### LINKED ISSUE #253: sql`${sql.join(sql.binary(…))}` results in unexpected value expression.
maybe this isn't a bug and i've misunderstood how something to be used. and either way i've worked around it so this is not pressing.

but while this works:
```
sql`${sql.binary(Buffer.from('test'))}`
```

and even this:
```
sql.join([sql.binary(Buffer.from('test'))], sql``)
```

this results in `Unexpected value expression`:
```
sql`${sql.join([sql.binary(Buffer.from('test'))], sql``)}`
```

szist:
fixes #253


gajus:
Thank you

PR #313, opened 2021-11-24
preserve ssl options

SychevSP:
The library allows to specify custom ssl options like this:
```
const ca = readFileSync('/etc/ssl/ca.crt', 'utf8');
export const pool = createPool(pgUrl, {ssl: {ca},});
```
However these ssl options are overridden:
https://github.com/gajus/slonik/blob/master/src/factories/createPoolConfiguration.ts#L29
 
So it seams the library discards user provided ssl options like custom certificate authority. 

In this pull request ssl options are preserved


gajus:
This was an oversight in implementation. f18f1b7 adds the desired behaviour.

PR #322, opened 2021-12-30
Added config parameter to stream with batch size

bdbernardy:
The `pool.stream` method is prohibitively slow when working with large amounts of data because the QueryStream is configured to fetch 100 rows at a time. Being able to control the number of rows fetched would allow us to find the right balance between execution speed and memory consumption.

This PR adds a config parameter to the `pool.stream` method which allows us to specify the batchSize used in the QueryStream.

### DIFF
--- a/src/QueryStream.ts
+++ b/src/QueryStream.ts
@@ -38,7 +38,7 @@ export class QueryStream extends Readable {
 
   public handleError: Function;
 
-  public constructor (text: unknown, values: unknown, options?: ReadableOptions & {batchSize: number, }) {
+  public constructor (text: unknown, values: unknown, options?: ReadableOptions & {batchSize?: number, }) {
     super({
       objectMode: true,
       ...options,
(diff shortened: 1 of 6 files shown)


gajus:
Thank you

bdbernardy:
Awesome! Thank you. I just tested it and it works wonders. Streaming 2 million rows now only takes 6 minutes instead of 2 hours :).

gajus:
yay, thank you for contributing

raine:
Is `batchSize` supposed to control how many rows you get in the `data` event? Because I'm getting a single one even with `batchSize: 1000`. Or is it some internal thing?

bdbernardy:
> Is `batchSize` supposed to control how many rows you get in the `data` event? Because I'm getting a single one even with `batchSize: 1000`. Or is it some internal thing?

It's the number of rows that you fetch from the database at a time. The higher the number, the faster it will be, but it will also increase the memory consumption.

PR #323, opened 2022-01-11
types: replace @ts-expect-error with type cast

npdev453:


gajus:
Superseded by https://github.com/gajus/slonik/pull/337

PR #329, opened 2022-02-09
Fix readme example using incorrect require

mull:


gajus:
These changes need to be done in `.README`

mull:
Oops I didn't notice how the README was put together. Fixed!

PR #333, opened 2022-03-26
feat: add jsonb (fixes #317)

### LINKED ISSUE #317: `sql.json(null)` should produce `'null'::jsonb` instead of `null`
## Expected Behavior
The query
```ts
sql`
  SELECT ${sql.json(null)}
`;
```
should result in the SQL
```sql
SELECT 'null'::jsonb
```

## Current Behavior
The above query actually gets translated to
```sql
SELECT NULL
```
`NULL` and `'null'::jsonb` are different, and this can have a bad side effect such as when using `jsonb_set`. In the following example, we have a jsonb `{"key": "value"}`, and try to set the value to `null`, which should result in a new jsonb `{"key": null}`
```ts
sql`
  SELECT jsonb_set('{"key": "value"}'::jsonb, '{key}', ${sql.json(null)})
`
```
However, it actually results in `NULL` instead of `{"key": null}` since Slonik passes in `NULL` instead of `'null'::jsonb`. Note that you could also simply pass in the string `'null'` and it would be implicitly casted to jsonb here.

## Possible Solution
I haven't followed the code path for how `sql.json` gets translated, but I did find [a test case that is specifically checking `null`](https://github.com/gajus/slonik/blob/da62a645715e907c24ec6bc0837d80e45bb5be8d/test/slonik/templateTags/sql/json.ts#L42) but expecting what I consider an incorrect result.
The test is:
```ts
test('passes null unstringified', (t) => {
  const query = sql`SELECT ${sql.json(null)}`;

  t.deepEqual(query, {
    sql: 'SELECT $1',
    type: SqlToken,
    values: [
      null,
    ],
  });
});
```
but I believe it should be:
```ts
test('passes null unstringified', (t) => {
  const query = sql`SELECT ${sql.json(null)}`;

  t.deepEqual(query, {
    sql: 'SELECT $1',
    type: SqlToken,
    values: [
      'null',
    ],
  });
});
```

## Workaround
I ran into this issue when trying to do something like
```ts
sql`SELECT jsonb_set(data, path, ${sql.join(value)}`
```
where `value` can sometimes be null. Instead I've rewritten this as:
```ts
sql`SELECT jsonb_set(data, path, ${value === null ? sql`'null'::jsonb` : sql.json(value)})`
```
The ternary works, but it would be simpler if `sql.json` handled `null` differently.

gajus:
* changes `sql.json` to produce bindings with `::json` type identifier
* adds new `sql.jsonb` method
* `sql.json` is now `'null':json` (previously it was `NULL`)

If you desire the old `sql.json` `NULL` behavior, you will need to refactor code to:

```ts
const foo = /* maybe NULL */;

sql`
SELECT ${foo === null ? NULL : sql.json(foo)}
`
```

BREAKING CHANGE:

* previously `sql.json` did not specify `::json` or `::jsonb`. Now the appropriate type identifier is chosen based on whether you use `sql.json` or `sql.jsonb`. In future, this may be leveraged to achieve greater performance when using `sql.jsonb`.


.README/QUERY_BUILDING.md
```diff
@@ -172,20 +172,43 @@ Produces:
...
+  value: SerializableValue
+) => JsonBinarySqlToken;
+
+```
+
+Serializes value and binds it as a JSON binary, e.g.
+
+```js
+await connection.query(sql`
+  SELECT (${sql.json([1, 2, 3])})
```
nponiros, line 196: I believe that this should be `jsonb` instead of `json`

thecodeboss (approved):
Looks solid to me, thanks for tackling this. Tests look great too.

PR #340, opened 2022-04-08
feat: add idle timeout (fixes #269)

### LINKED ISSUE #269: idleTimeout (option for createPool) doesn't accept  'DISABLE_TIMEOUT'
## Expected Behavior

According to the docs, we can use
```
createPool(pgUrl, { idleTimeout: 'DISABLE_TIMEOUT' })
```

## Current Behavior

In https://github.com/gajus/slonik/blob/master/src/factories/createPoolConfiguration.ts we see that the `'DISABLE_TIMEOUT'` string is not used anywhere for `idleTimeout` (it's used only for `connectionTimeout`).

## Possible Solution

Maybe do something similar to what is done in `connectionTimeout`?
```
poolConfiguration.idleTimeoutMillis = undefined
```

It's not clear if this is supported in `pg`: https://node-postgres.com/api/pool

gajus:


src/factories/createPoolConfiguration.ts
```diff
@@ -58,6 +58,16 @@ export const createPoolConfiguration = (dsn: string, clientConfiguration: Client
...
+  if (clientConfiguration.idleTimeout !== 'DISABLE_TIMEOUT') {
+    if (clientConfiguration.idleTimeout === 0) {
+      log.warn('idleTimeout=0 sets timeout to 0 milliseconds; use idleTimeout=DISABLE_TIMEOUT to disable timeout');
+
+      poolConfiguration.idleTimeoutMillis = 1;
+    } else {
+      poolConfiguration.idleTimeoutMillis = clientConfiguration.idleTimeout;
+    }
+  }
+
```
AureaMohammadAlavi, line 70: If `idleTimeoutMillis` option is not set, `pg-pool` sets this to 10,000 milliseconds.
https://github.com/brianc/node-postgres/blob/a4ef6ce38c1e04bad2215312b1c79e64654cc857/packages/pg-pool/index.js#L92

If `clientConfiguration.idleTimeout` is set to `DISABLE_TIMEOUT`, then `poolConfiguration.idleTimeoutMillis` should be set to 0 to disable auto-disconnection of idle clients.

PR #345, opened 2022-05-06
Update README.md

nairihar:
Fix example parameter


gajus:
Fixed in https://github.com/gajus/slonik/commit/451cbad06eccca3f95a319f8cc0b1720e35fc2db

PR #356, opened 2022-07-20
add unimplemented note to createSqlTag in readme

qpwo:
I was trying to insert bigint's and was getting `Unexpected value expression`. Then I realized `createSqlTag` doesn't actually take `typeParsers`.

Would actually implement it in a PR myself if I had the expertise & spare time

### DIFF
--- a/README.md
+++ b/README.md
@@ -1223,7 +1223,7 @@ import {
 
 ```
 
-Sometimes it may be desirable to construct a custom instance of `sql` tag. In those cases, you can use the `createSqlTag` factory, e.g.
+**(unimplemented)** Sometimes it may be desirable to construct a custom instance of `sql` tag. In those cases, you can use the `createSqlTag` factory, e.g.
 
 ```js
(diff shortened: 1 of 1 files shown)


qpwo:
Could also note that some values can still be put in an sql template string if you make them strings. E.g.
```
sql`INSERT INTO foo VALUES (${someBigInt.toString()});`
```

gajus:
Not clear what the author wants. Documentation is accurate as is.

qpwo:
createSqlTag takes no arguments: https://github.com/gajus/slonik/blob/1561e6205362eafa00a3bdc23b917a5f01a5c51d/src/factories/createSqlTag.ts#L189

Readme

```ts
/**
 * @param {SqlTagConfiguration} configuration
 */
const sql = createSqlTag(configuration);
```

But passing the configuration actually does nothing¿

gajus:
Misread. Removed the reference.

PR #360, opened 2022-07-30
Close underlying QueryStream on stream.destroy()

### LINKED ISSUE #347: Closing cursor after destroying stream
Using v28.1.0, I'm observing that in some cases after I destroy a stream, I'm unable to use the connection to then run a different query. If `foo` is a large table, then the following script never completes:

```js
const { createPool, sql } = require('slonik');

// Specify your connection URI here.
const pool = createPool('postgres://');

const main = () => pool.connect(async (connection) => {
  await connection.stream(sql`SELECT * FROM foo`, (stream) => {
    stream.destroy();
    console.log('stream destroyed');
  });
  console.log('before query');
  await connection.query(sql`SELECT COUNT(*) FROM bar`);
  console.log('all done');
});

main();
```

I see that `stream destroyed` and `before query` are logged, but `all done` is not: it seems that the promise that `connection.query()` returns never settles.

Note that `foo` must be ~a large table~ a table of a certain size. If the `connection.stream()` query returns fewer rows, then I don't see this behavior. For example, if I make the following change, then I don't see this behavior:

```diff
-  connection.stream(sql`SELECT * FROM foo`, (s) => {
+  connection.stream(sql`SELECT COUNT(*) FROM foo`, (s) => {
```

After a little investigation, I'm thinking that the cause of the behavior is that calling `stream.destroy()` doesn't close the underlying `QueryStream`. If the `QueryStream` isn't closed, then I think the `pg-cursor` `Cursor` also won't be closed.

If I add logging to the `close()` method of `QueryStream`, then I don't see `close()` being called when I run the script above. But if I modify `stream()` to call `queryStream.close()` when `transformedStream` is closed below, then the script completes successfully:

https://github.com/gajus/slonik/blob/371f15257451ee5c918e817aba7879a4c04f20b8/src/connectionMethods/stream.ts#L68-L77

alxndrsn:
Fixes #347

### DIFF
--- a/src/QueryStream.ts
+++ b/src/QueryStream.ts
@@ -63,15 +63,10 @@ export class QueryStream extends Readable {
     this.cursor.submit(connection);
   }
 
-  public close (callback: Function) {
+  public _destroy (error: Error, callback: Function) {
     this._closed = true;
 
-    // eslint-disable-next-line unicorn/consistent-function-scoping
-    const close = () => {
-      this.emit('close');
-    };
-
-    this.cursor.close(callback || close);
+    this.cursor.close(callback);
   }
 
   public _read (size: number) {
--- a/src/connectionMethods/stream.ts
+++ b/src/connectionMethods/stream.ts
@@ -1,4 +1,6 @@
-import type Stream from 'stream';
+import {
+  type Readable,
+} from 'stream';
 import through from 'through2';
 import {
   QueryStream,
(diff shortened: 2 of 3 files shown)


alxndrsn:
I've seen calling `stream.destroy(new Error())` affected by https://github.com/brianc/node-postgres/issues/2556, which has been fixed in `pg` but not released yet.

alxndrsn:
@gajus I've rebased this onto the current `main` (`v30.4.0`).  Without the accompanying changes in `src/`, the tests in this PR time out.

gajus:
Thank you

PR #365, opened 2022-08-03
feat: use zod to validate query output and infer query types (fixes #364)

### LINKED ISSUE #364: integrate zod into Slonik

```ts
  SELECT id, name
  FROM person
`);
```

```ts
  ${z.object({
    id: z.number(),
    name: z.string(),
  })}
  SELECT id, name
  FROM person
`);
```

```ts
  ${z.object({
    id: z.number(),
    name: z.string(),
  })}
  SELECT id, name, description
  FROM person
`);
```

gajus:


src/factories/createSqlTag.ts
```diff
@@ -47,8 +50,20 @@ const log = Logger.child({
 
 const sql: SqlTaggedTemplate = (
   parts: readonly string[],
-  ...values: readonly ValueExpression[]
+  ...args: readonly [ValueExpression | ZodTypeAny, ...ValueExpression[]]
```
mmkal, line 53: Question re this api design. It's a little confusing having a zod object seemingly randomly in there with the query, and opens the door to incorrect stuff like

```ts
sql`
select ${z.object({ foo: z.string() })}, 123
```

could a similar goal be achieved with something like

```ts
const getSql = <T>(type: z.ZodType<T>) => {
  return (...args: ValueExpression[]) => ({
    ...sql<T>(...args),
    zodObject: type,
  })
}
```

then if you do something like `sql.type = getSql` you could use like:

```ts
const foo = sql.type(z.object({ foo: z.string() }))`
  select 'abc' as foo
`
```

Or if the type is big:

```ts
const Foo = z.object({ foo: z.string() })

const foo = sql.type(Foo)`
  select 'abc' as foo
`
```

that would also allow for other validation libraries that confirmed to Zod's interface. Wouldn't need to do instanceof type checks.

gajus: This makes sense.

gajus: After a bit of testing, going this direction. Thanks @mmkal !

gajus:
Will start a new PR.

PR #367, opened 2022-08-04
zod validation (fixes #364)

### LINKED ISSUE #364: integrate zod into Slonik

gajus:


test/types.ts
```diff
@@ -4,148 +4,86 @@
  */
 
 import {
-  expectTypeOf,
-} from 'expect-type';
+  expectType,
+} from 'tsd';
```
mmkal, line 8: I'm the author of expect-type and not tsd, so I might be biased (although both are free!) - but I think a switch to tsd should be done _very_ carefully if dealing with generics in any way because of this bug: https://github.com/SamVerschueren/tsd/issues/142

IMO it's not worth the risk. It's a serious bug for a type assertion library because it'll tell you everything is fine even when it's not. expect-type doesn't have that problem (and I think the assertions are a little more readable).

test/types.ts
```diff
@@ -4,148 +4,86 @@
...
   const any = await client.any(sql``);
-  expectTypeOf(any).toEqualTypeOf<ReadonlyArray<Record<string, PrimitiveValueExpression>>>();
+  expectType<readonly any[]>(any);
 
-  const anyTyped = await client.any<Row>(sql``);
-  expectTypeOf(anyTyped).toEqualTypeOf<readonly Row[]>();
-
-  const anyTypedQuery = await client.any(sql<Row>``);
-  expectTypeOf(anyTypedQuery).toEqualTypeOf<readonly Row[]>();
+  const anyTyped = await client.any(sql.type(Row)``);
```
mmkal, line 31: I don't think the old test should be deleted? Is slonik going to stop supporting `` sql<Row>`select 123` ``?? I would have thought this is just a new feature and should have corresponding new type tests. Right now this PR deletes many many tests that ensure slonik's types are correct for end users (since test/dts.ts also got deleted entirely).

mmkal (commented):
Pretty scared by the deletion of the type tests. I wrote most of them for very specific use cases that my team/I have in production in multiple projects. Is slonik no longer supporting those use cases?

test/types.ts
gajus: I don't see a path for supporting both. They don't mix will.

What happens if someone does:

```ts
sql<{id: number}>.type(
    z.object({
      id: z.string()
    })
)`select 123`
```

test/types.ts
gajus: Fair enough – re-added `expect-type`.

test/types.ts
mmkal: Not a risk - that isn't valid ts syntax. The generic on `sql` is only when used as a function directly. `sql.type` is another, independent, generic function which returns a non-generic function. I'm very sure of this, and still pretty concerned about abandoning the existing (and totally satisfactory!) form slonik has now. If you're not convinced, can I have a few days to come up with a pull request which proves this out?

gajus: > If you're not convinced, can I have a few days to come up with a pull request which proves this out?

Absolutely!

gajus:
Closing in favor of https://github.com/gajus/slonik/pull/369.

PR #369, opened 2022-08-04
feat: add zod validation

gajus:
This PR most likely supersedes https://github.com/gajus/slonik/pull/367


src/factories/createSqlTag.ts
```diff
@@ -92,19 +96,62 @@ const sql: SqlTaggedTemplate = (
...
     configurable: false,
     enumerable: true,
     writable: false,
   });
 
-  return query;
+  return query as unknown as SqlSqlToken<QueryResultRow>;
+};
+
+sql.type = <T extends ZodTypeAny>(zodObject: T) => {
```
mmkal, line 129: What do you think about keeping it decoupled from zod, and just taking a parser in a shape that zod conforms to? Like 

```suggestion
interface Parser<T> {
  parse(input: unknown): T
}

sql.type = <T>(parser: Parser<T>) => {
```

No real benefit to explicitly coupling to zod I don't think.

src/factories/createSqlTag.ts
```diff
@@ -92,19 +96,62 @@ const sql: SqlTaggedTemplate = (
...
+      zodObject: zodObject.strict(),
+    };
+
+    Object.defineProperty(query, 'sql', {
+      configurable: false,
+      enumerable: true,
+      writable: false,
+    });
+
+    return query as unknown as SqlSqlToken<T>;
```
mmkal, line 153: Shouldn't this be something like 
```suggestion
    return query as unknown as SqlSqlToken<z.infer<T>>;
```

Or if you take the Parser<T> suggestion this line would be correct as is

src/types.ts
```diff
@@ -409,49 +409,50 @@ export type InternalNestedTransactionFunction = <T>(
   transactionRetryLimit?: number,
 ) => Promise<T>;
 
-// eslint-disable-next-line @typescript-eslint/no-empty-interface, @typescript-eslint/consistent-type-definitions, @typescript-eslint/no-unused-vars
-export interface TaggedTemplateLiteralInvocation<Result extends UserQueryResultRow = QueryResultRow> extends SqlSqlToken { }
+type MixedRow = QueryResultRow | ZodTypeAny;
```
mmkal, line 412: A union of two totally distinct types for such a core piece of the library, and so tied up in zod's abstractions doesn't feel right to me. I think either of my previous two suggestions would remove the need for this big change. I really don't think a breaking change is necessary!

mmkal (commented):
I like that this doesn't necessarily affect end-users - but I think the zod types can be abstracted away more.

gajus:
Happy to accept it as an improvement, but I don't necessarily see major benefit to loose coupling. Zod is a leading choice for this purpose.

src/routines/executeQuery.ts
```diff
@@ -346,6 +347,26 @@ export const executeQuery = async (
 
   // Stream does not have `rows` in the result object and all rows are already transformed.
   if (result.rows) {
+    if (slonikSqlRename.zodObject && slonikSqlRename.zodObject?._def?.typeName !== 'ZodAny') {
+      for (const row of result.rows) {
+        const validationResult = slonikSqlRename.zodObject.safeParse(row);
```
mmkal, line 352: Zod objects can transform data, so this might lead to unexpected behaviour. e.g. an object like `z.object({ foo: z.string().transform(s => s.split(',')) })` has an "output" type of `{ foo: string[] }` but since all this does is validate (it ignores the successful parse result), the row type at runtime will be `{ foo: string }`, and the code will expect `{ foo: string[] }`. An `interceptor.transformRow` value could be dynamically added to properly handle this, maybe?

mmkal (commented):
I'm working on a hopefully less invasive implementation of this, that doesn't affect the existing types or functionality.

PR #371, opened 2022-08-05
feat: zod parser, non-breaking

### LINKED ISSUE #364: integrate zod into Slonik

mmkal:
This is a fully backwards-compatible, non-breaking iteration on #369 

Fixes #364 (although @gajus I think the discussion on #364 should be resolved before merging, and I'd be interested to get @janpaepke's take on this too - if you'd like to integrate this into slonik-typegen, we are the two people to get on board!)

Essentially, this adds a `parser` property to the `SqlSqlToken` type - which instructs slonik to parse the query result rows before returning them. There's a _bias_ towards zod, and in docs zod should be mentioned as the only type parsing solution that will be officially supported, but there's no reason to leak Zod's abstractions into slonik explicitly.

All tests from #369 have been included here and are passing, and the tests that were deleted in #369 have been restored and are also passing - meaning this is a backwards compatible change.

The one exception to this is that `.strict()` is no longer called internally - though of course it can be called by the creator of the zod object - and one of the tests is updated to reflect this. See https://github.com/gajus/slonik/pull/369/files#r938576111

https://github.com/gajus/slonik/pull/369/files#r938624301 is still an issue, I will add a failing test to prove this and (hopefully) a fix too.

### DIFF
--- a/src/routines/executeQuery.ts
+++ b/src/routines/executeQuery.ts
@@ -347,9 +347,12 @@ export const executeQuery = async (
 
   // Stream does not have `rows` in the result object and all rows are already transformed.
   if (result.rows) {
-    if (slonikSqlRename.parser) {
-      for (const row of result.rows) {
-        const validationResult = slonikSqlRename.parser.safeParse(row);
+    const {parser} = slonikSqlRename
+
+    // if a parser exists, use it to transform the row. This allows for zod objects like `z.object({ foo: z.string().transform(s => s.split(',') )})`
+    const parserInterceptor: typeof clientConfiguration.interceptors[0] | undefined = parser && {
+      transformRow: (_executionContext, _actualQuery, row) => {
+        const validationResult = parser.safeParse(row);
 
         if (!validationResult.success) {
           log.error({
@@ -364,9 +367,15 @@ export const executeQuery = async (
             validationResult.error.issues,
           );
         }
+        return validationResult.data
       }
     }
-    for (const interceptor of clientConfiguration.interceptors) {
+
+    const interceptors = parserInterceptor
+      ? [...clientConfiguration.interceptors, parserInterceptor]
+      : clientConfiguration.interceptors
+
+    for (const interceptor of interceptors) {
       if (interceptor.transformRow) {
         const {
           transformRow,
--- a/src/routines/executeQuery.ts
+++ b/src/routines/executeQuery.ts
@@ -22,6 +22,7 @@ import {
   UnexpectedStateError,
   UniqueIntegrityConstraintViolationError,
   TupleMovedToAnotherPartitionError,
+  SchemaValidationError,
 } from '../errors';
 import {
   getPoolClientState,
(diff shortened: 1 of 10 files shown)


src/routines/executeQuery.ts
```diff
@@ -346,7 +347,35 @@ export const executeQuery = async (
...
+            validationResult.error.issues,
+          );
+        }
+        return validationResult.data
+      }
+    }
+
+    const interceptors = parserInterceptor
+      ? [...clientConfiguration.interceptors, parserInterceptor]
+      : clientConfiguration.interceptors
```
mmkal, line 376: fix for https://github.com/gajus/slonik/pull/369/files#r938624301

gajus:
Looks good!

> The one exception to this is that .strict() is no longer called internally - though of course it can be called by the creator of the zod object - and one of the tests is updated to reflect this. See https://github.com/gajus/slonik/pull/369/files#r938576111

Need to find a way how to force `strict()`. Users should be able to bypass type validation by simply omitting properties.

mmkal:
> Looks good!
> 
> > The one exception to this is that .strict() is no longer called internally - though of course it can be called by the creator of the zod object - and one of the tests is updated to reflect this. See https://github.com/gajus/slonik/pull/369/files#r938576111
> 
> Need to find a way how to force `strict()`. Users should be able to bypass type validation by simply omitting properties.

I **strongly** disagree. `.strict()` is the domain of zod, not of slonik. It's creating an unnecessary coupling with zod that will confuse, inconvenience, and cause unnecessary errors. Here are some reasons:

1. `.strict()` assumes a `ZodObject` which we otherwise wouldn't need to assume
2. `.strict()` might error or behave unexpectedly if the type passed in is *already* strict, which it might be depending on the preferences of the end-user
3. `.strict()` will throw errors at runtime which the user might not have expected. So unless there is *perfect* test coverage (which there is for approximately zero real-world projects in existence) we may be causing production outages for downstream users.

On top of that, it's just not necessary. See [ca06fe6](https://github.com/gajus/slonik/pull/371/commits/ca06fe623891e20b57f93ba69f2a819cd54b5679) for an example of a test that shows the application _won't_ be able to use to undefined keys, just by relying on the default behaviour of zod. There's no need for slonik to overstep its boundaries like it would be with .strict().

See also [5e3ae4d](https://github.com/gajus/slonik/pull/371/commits/5e3ae4d2f32cd281679dce9b94fce2d60609d745) for an example of why else it's not necessary - if you want to use `.strict()` at your company, you of course can - just use `.strict()` on the values passed into `sql.type(...)`. I'd be happy to chat about how to achieve this in slonik-typegen, if you'd like.

gajus:
> .strict() assumes a ZodObject which we otherwise wouldn't need to assume

I don't see the benefit of not assuming `ZodObject`.

Using loose types that just cover our needs to avoid coupling with Zod – that's fine.

However, compromising on what the library offers just to avoid coupling is pushing it.

I don't expect anyone to use Slonik with non-Zod parser.

> .strict() might error or behave unexpectedly if the type passed in is already strict, which it might be depending on the preferences of the end-user

There is a guard for that

https://github.com/gajus/slonik/pull/375/files#diff-a537163e5406039ab7b8e6758fb354f87daf163664dce8a6d881d04c62473ceaR115-R117

> .strict() will throw errors at runtime which the user might not have expected. So unless there is perfect test coverage (which there is for approximately zero real-world projects in existence) we may be causing production outages for downstream users.

The whole point of adding runtime validation is to catch if/when your types are not aligned with what the query returns.

gajus:
> On top of that, it's just not necessary. See [ca06fe6](https://github.com/gajus/slonik/pull/371/commits/ca06fe623891e20b57f93ba69f2a819cd54b5679) for an example of a test that shows the application won't be able to use to undefined keys, just by relying on the default behaviour of zod. There's no need for slonik to overstep its boundaries like it would be with .strict().

While it is not strictly necessary (application can only use defined types), it allows writing code that might be pulling unnecessary data.

```ts
sql.type(z.({id: z.string()}))`
SELECT
  id,
  password
FROM user
`
```

This should not be encouraged for hopefully obvious reasons.

PR #377, opened 2022-08-08
Failing test: should be able to destroy a stream with an error

alxndrsn:


test/slonik/integration/pg.ts
```diff
@@ -337,3 +337,21 @@ test('copies from binary stream', async (t) => {
...
+test('does not crash after destroying a stream with an error', async (t) => {
+  const pool = await createPool(t.context.dsn);
+
+  await pool.stream(sql`
+    SELECT * FROM GENERATE_SERIES(1, 100)
+  `, (stream) => {
+    stream.destroy(new Error('Foo'));
+  });
+
+  t.deepEqual(await pool.anyFirst(sql`
```
alxndrsn, line 350: Perhaps this should test that expected events were emitted:

> Destroy the stream. Optionally emit an 'error' event, and emit a 'close' event (unless emitClose is set to false). After this call, the readable stream will release any internal resources and subsequent calls to push() will be ignored.
> -- _https://nodejs.org/api/stream.html#readabledestroyerror_

PR #392, opened 2022-08-30
Update VS code SQL highlighting extension

aust1nz:
This is a simple readme-only update that points users to the [vscode-sql-lit](https://marketplace.visualstudio.com/items?itemName=thebearingedge.vscode-sql-lit) extension.

The extension was recently updated and supports SQL highlighting even when using the ``sql.type(zodType)`SELECT * from Users`;`` syntax, unlike the previously recommended extension.

### DIFF
--- a/.README/README.md
+++ b/.README/README.md
@@ -99,7 +99,7 @@ For more information, refer to the [JavaScript Tagged Template Literal Grammar E
 
 ### VS Code Syntax Highlighting Extension
 
-The [`vscode-sql-template-literal` extension](https://marketplace.visualstudio.com/items?itemName=forbeslindesay.vscode-sql-template-literal) provides syntax highlighting for VS Code:
+The [`vscode-sql-lit` extension](https://marketplace.visualstudio.com/items?itemName=thebearingedge.vscode-sql-lit) provides syntax highlighting for VS Code:
 ![Syntax highlighting in VS Code](./.README/vscode-syntax-highlighting.png)
 
 ## Development


gajus:
Nice, I was looking for something that would work with `.type`!

gajus:
@aust1nz are you the author of lit?

Unfortunately it does not work if `sql.type` definition is multi-line, e.g.

```ts
sql.type(
  z.object({
    id: z.number(),
    // ...
  }),
)`
  SELECT
    tr1.id,
    tr1.created_at "createdAt",
    tr1.started_at "startedAt"
`
```

aust1nz:
Ah, didn't look for that syntax! I'm not the library maintainer but did open a PR last week that the maintainer merged/pushed. Let me see if I can figure out the appropriate regex for that syntax. I'll follow up if I can git the lib updated.

aust1nz:
Ack, upon further review, it doesn't look like multi-line syntax highlighting is supported in VS Code. [Source](https://github.com/microsoft/vscode-textmate/issues/41#issuecomment-344294298)

So, neither library will highlight this syntax properly:

```ts
sql.type(
  z.object({
    id: z.number(),
    // ...
  }),
)`
  SELECT
    tr1.id,
    tr1.created_at "createdAt",
    tr1.started_at "startedAt"
`
```

However, the suggested vscode-sql-lit library IS an improvement over the current selection, because it WILL successfully highlight the following:

```ts
const zodObject =
  z.object({
    id: z.number(),
    // ...
  });

sql.type(zodObject)`
  SELECT
    tr1.id,
    tr1.created_at "createdAt",
    tr1.started_at "startedAt"
`

Defer to your judgment if that level of improvement merits a updated docs.

PR #401, opened 2022-09-07
Adopt MixedRow

### LINKED ISSUE #397: Cannot infer parser type
## Expected Behavior

```ts
const parser = sql.type(ZodRow)``.parser;
expectTypeOf(parser).toEqualTypeOf<z.infer<typeof ZodRow>>();
```

should pass.

## Current Behavior

`parser` type is `ParserUserQueryRow`.

### LINKED ISSUE #398: Cannot extend parser
## Expected Behavior

```ts
const personQuery = sql.type(
  z.object({
    id: z.number(),
    name: z.string(),
  })
)`
  SELECT id, name
  FROM person
`;

const personWithRelationship = sql.type(
  personQuery.parser.extend({
    relationship: z.string(),
  })
)`
  SELECT t1.*, 'friend' relationship
  FROM ${personQuery} t1
`;
```

Expected this to work.

## Current Behavior

Parser does not have `extend` interface.

## Possible Solution

Adopt `Zod` object as the parser.

gajus:
BREAKING CHANGE:

* drops `ParserIssue` and `Parser<T>`, which were Zod-like types in favor of using Zod
* drops `.strict()`
* drops the ability to hint return type using an arbitrary object (use `sql.type` instead)
* drops the ability to hint return type using interfaces (use `sql.type` instead)
* exports `SchemaValidationError`
* fixes #398 
* fixes #397 

Context around the changes:

As we rolled out `sql.type` across our repositories in contra.com, it became clear that every use of `sql<T>` is an anti-pattern. Furthermore, the earlier implementation meant that we could not fully leverage the power of [Zod](https://github.com/colinhacks/zod) to extend and manipulate result types/shape (see #398 and #397). As such, this release is a stepping stone towards abandoning `sql<T>` in favor of `sql.type` – the next major release is most likely not going to support `sql<T>` and will instead double down on deeper Zod integration.

While this change might not be ideal for everyone, esp. if you have no desire to use Zod, based on my experience, I firmly believe that this is the best way forward to create the safest client for PostgreSQL.

The next release most likely to include these changes:

* `sql` will get renamed to `sql.fragment` – to separate query fragments from top-level queries.
* `sql` alone won't be available.
* `sql.fragment` is not going to be a valid input to query methods.
* `sql.type` will be the expected input of all query methods.
* `sql.unsafe<T>` is going to be added to address edge cases

CC @danielrearden @Boeing787 @mmkal

### DIFF
--- a/.README/RUNTIME_VALIDATION.md
+++ b/.README/RUNTIME_VALIDATION.md
@@ -107,9 +107,9 @@ You can infer the TypeScript type of the query result. There are couple of ways 
 // Infer using z.infer<typeof yourSchema>
 // https://github.com/colinhacks/zod#type-inference
 type Person = z.infer<typeof personObject>;
-// from sql tagged template `zodObject` property
+// from sql tagged template `parser` property
 type Person = z.infer<
-  personQuery.zodObject
+  personQuery.parser
 >;
 ```
 
--- a/README.md
+++ b/README.md
@@ -1322,9 +1322,9 @@ You can infer the TypeScript type of the query result. There are couple of ways 
 // Infer using z.infer<typeof yourSchema>
 // https://github.com/colinhacks/zod#type-inference
 type Person = z.infer<typeof personObject>;
-// from sql tagged template `zodObject` property
+// from sql tagged template `parser` property
 type Person = z.infer<
-  personQuery.zodObject
+  personQuery.parser
 >;
 ```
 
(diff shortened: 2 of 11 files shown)


mgreenw:
@gajus I built a thing wrapper around `slonik` that uses `zod` for runtime validation and it's been awesome. I completely agree with your points above and I'm so excited that you've now built this in to `slonik`!

gajus:
@mgreenw I will legit never look at development the same way after this discovery.

It completely changed my thinking how a contract between a codebase and database should look.

Lots of exciting things coming.

You may also find this useful https://github.com/gajus/slonik/issues/408

PR #412, opened 2022-09-19
QueryStream: destroy() on error

alxndrsn:
I think this would be a more correct way to handle errors when reading from streams.

From https://nodejs.org/api/stream.html#stream_errors_while_reading:

> Errors occurring during processing of the [`readable._read()`](https://nodejs.org/api/stream.html#readable_readsize) must be propagated through the [`readable.destroy(err)`](https://nodejs.org/api/stream.html#readable_destroyerr-callback) method. Throwing an `Error` from within [`readable._read()`](https://nodejs.org/api/stream.html#readable_readsize) or manually emitting an `'error'` event results in undefined behavior.


alxndrsn:
This seems correct from the quoted documentation, but has unexpected effects on error propagation when creating a stream with bad SQL syntax.

alxndrsn:
@gajus thanks for updating this!  Is there anything I can do to improve the PR?

gajus:
Thank you

Issue #418, opened 2022-10-05
Cannot re-use pool connection after stream-related error


alxndrsn:
## Expected Behavior

Any connection returned to the connection pool should still work.

## Current Behavior

In certain circumstances, re-using a pool connection which has previously been used with `.stream()` and had a specific Error will throw when re-used.

## Possible Solution

I think the underlying bug is a race condition in `pg`.  I'm looking into a lower-level fix there.  See: https://github.com/brianc/node-postgres/issues/1674

## Steps to Reproduce

PR containing failing test at: https://github.com/gajus/slonik/pull/417

## Logs

I've seen a couple of different failures, e.g.:

### Failure 1

```js
  ✖ integration › pg › can re-use pool after stream() and connection timeout

  test/slonik/integration/pg.ts:409

   408:                                       
   409:   t.deepEqual(await pool.anyFirst(sql`
   410:     SELECT TRUE                       

  Difference:

    [
  +   true,
    ]

  › test/slonik/integration/pg.ts:409:5

  ─

  1 test failed
```

### Failure 2

```ts
  ✖ integration › pg › can re-use pool after stream() and connection timeout 

  Uncaught exception in test/slonik/integration/pg.ts

  TypeError: Cannot read properties of null (reading 'handleRowDescription')

  › Client._handleRowDescription (node_modules/pg/lib/client.js:340:22)
  › node_modules/pg/lib/connection.js:114:12
  › Parser.parse (node_modules/pg-protocol/src/parser.ts:104:9)
  › Socket.<anonymous> (node_modules/pg-protocol/src/index.ts:7:48)

  ✖ test/slonik/integration/pg.ts exited with a non-zero exit code: 1
```

PR #419, opened 2022-10-05
Allow re-use of pool connection after stream() and Error

alxndrsn:
Hack to improve behaviour around #418


gajus:
Closing as these PRs never passed the checks.

PR #420, opened 2022-10-05
pool: don't allow connection re-use after stream()

alxndrsn:
Hack to improve behaviour around https://github.com/gajus/slonik/issues/418

Workaround for https://github.com/brianc/node-postgres/issues/1674

Should be more robust that #419


alxndrsn:
This is failing `tsc`, but I'm not sure how to resolve the errors there.  Tests are passing.

PR #424, opened 2022-10-20
Remove baked in result parser in favor of a middleware

gajus:


gajus:
CC @mmkal 

This does not fully address your concerns, since you are still required to use Zod to instrument query types, but it may unblock you in certain areas.

Issue #432, opened 2022-11-14
Slonik not compiling because of Record key typing error


MiK546:
I'm having trouble with the latest versions of Slonik not compiling at all even in an empty project I created just to test this out. The TypeScript compilation error I receive is:

```
> npx tsc
node_modules/slonik/dist/src/index.d.ts:25:24 - error TS2344: Type 'unknown' does not satisfy the constraint 'string | number | symbol'.

25         parser: Record<unknown, import("zod").ZodTypeAny>[Y];
                          ~~~~~~~


Found 1 error in node_modules/slonik/dist/src/index.d.ts:25
```

## Expected Behavior
The project compiles without errors.

## Current Behavior
The project does not compile because of TypeScript errors.

## Possible Solution
This error seems to be caused because `unknown` is used as the key of the record. If for example the `PropertyKey` type was used instead there would be no problem.
```ts
typeAlias: <Y extends PropertyKey>(parserAlias: Y) => (parts: readonly string[], ...args: readonly import("./types").ValueExpression[]) => Readonly<{
        parser: Record<PropertyKey, import("zod").ZodTypeAny>[Y];
```

I don't know if that would make sense otherwise though.

## Steps to Reproduce
Link to a code sandbox, hope it works (or more like doesn't in this case) https://codesandbox.io/s/festive-cray-exw74t. The error should come up once you run `npm build` / `npx tsc`.

## Logs
n/a I think.

nathanchapman:
Confirmed that v33.0.4 resolves this compilation error. I think we can close this issue now @gajus @MiK546

gajus:
Thank you

PR #433, opened 2022-11-16
Fix createSqlTag typing

MiK546:
This is my attempt at fixing typing for `createSqlTag()`. See #432 for the bug I'm attempting to fix with this.

The current definition leaves the type of the key of the `typeAliases` object too wide by default, making it unknown and thus unusable as a Record key. This leads to a TypeScript compilation error. This pull request makes the key type extend `PropertyKey` instead making only valid Record key types assignable as the key type.

### DIFF
--- a/src/factories/createSqlTag.ts
+++ b/src/factories/createSqlTag.ts
@@ -109,7 +109,7 @@ const createFragment = (
 };
 
 export const createSqlTag = <
-  K extends keyof Z,
+  K extends PropertyKey,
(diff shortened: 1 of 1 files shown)


nathanchapman:
@gajus any way we can go ahead and get this merged in and released?

gajus:
Sorry, sometimes PRs get lost in the flood of notifications. Running tests and will merge if they pass.

gajus:
Thank you

PR #451, opened 2023-02-08
fix: make zod a peer dependency (fixes #449)

### LINKED ISSUE #449: Zod local version must be equal to slonik zod version
If the version of zod in the local package.json is greater than slonik zod version, typescript refuses to compile indicating that the internal zod types are different.

Example:
slonik - `zod: 3.19.2`
local (with yarn add) - `zod: 3.20.2`

Error:
```
// ...
Types of parameters 'input' and 'input' are incompatible.
        Type 'Zod.ParseInput' is not assignable to type 'import("**local_project**/node_modules/zod/lib/helpers/parseUtil").ParseInput'.
// ... same with different parameter names
```

## Expected Behavior
Code will compile indipendent of semver minor version, eg. `3.x.x`.

## Current Behavior
Code will not compile if versions don't match.

## Possible Solution

Maybe peer dependency? Don't really know what could fix it, sorry.

## Steps to Reproduce

1. `yarn install slonik zod` -- as of today, 15 january 2023 the default of zod is `3.20.2`
2.  `sql.type(zodObject)`

## Logs
No runtime usage

gajus:


dilame:
Perfect! Can't wait to check this fix out!

PR #492, opened 2023-06-19
readme: remove mention of pg-query-stream

mull:
It was merged into node-postgres and pg-query-stream is archived.


gajus:
Thank you

PR #496, opened 2023-07-06
Add support for infinity values to timestamp parsers

pablosproject:
I found it difficult working with `infinite` values in typescript in the timestamp field. This PR adds support for when the timestamp value is infinity.

Current implementation returns `NaN`, which is not the correct value for the timestamp.


gajus:
Thank you for flagging this. Heads up that your PR contains an error (both cases are returning negative).

Issue #471, opened 2023-03-14
Pool is missing error handling


valeneiko:
If existing pool connection is terminated by the server, PgPool will emit `error` event. Slonik does not register a listener for that event, so it results in a crash instead. It can happen in case of DB failover, or if someone runs `pg_terminate_backend(...)` query.

`PgPool` is `EventEmitter` which can emit `error` events, which by default crash node process: [docs](https://nodejs.org/docs/latest-v18.x/api/events.html#error-events).

## Expected Behavior
Error is logged and ignored.

## Current Behavior
If `PgPool` emits `error` event, node crashes : [docs](https://nodejs.org/docs/latest-v18.x/api/events.html#error-events).

## Possible Solution
- Immediately after a `pool` is created, register `error` event listener to log the error.
  ```ts
  // https://github.com/gajus/slonik/blob/main/src/factories/createPool.ts#L80-L85
  pool.on('error', (error) => {
    poolLog.error({
      error: serializeError(error),
    }, 'client error');
  });
  ```
- Do the same for the `setupClient`:
  ```ts
  // https://github.com/gajus/slonik/blob/main/src/factories/createPool.ts#L61-L69
  setupClient.on('error', (error) => {
    poolLog.error({
      error: serializeError(error),
    }, 'setup client error');
  });
  ```

## Steps to Reproduce
1. Create slonik pool, establish a few connections and wait
   ```ts
   const pool = await createPool(...);
   await Promise.all(new Array(10).fill(null).map(() => pool.connect(() => Promise.resolve())));
   await setTimeout(10000);
   ```
2. Run the SQL query to terminate all connections (or crash the DB)
   ```sql
   SELECT *, pg_terminate_backend(pid)
   FROM pg_stat_activity
   WHERE pid != pg_backend_pid()
   ```
3. Observe process started in step1 crashed

gajus:
I am not able to replicate this issue @valeneiko 

https://github.com/gajus/slonik/pull/507

Based on the description, I would expect this test to fail somehow, but it is passing as expected.

https://github.com/gajus/slonik/pull/507/files#diff-87a6820c951329f6ddf0055e0afd4f289f3be2a0bd147d7dccab62657e0d699bR175

valeneiko:
I have just tried this script (a copy of your test) and got the exception + crash:
```js
import { setTimeout as delay } from 'timers/promises';
import { createPool, sql } from 'slonik';

async function main() {
[...]

PR #507, opened 2023-09-26
test: handle terminated backend

gajus:
Attempt to replicate https://github.com/gajus/slonik/issues/471


gajus:
This is likely a non-issue.

Issue #338, opened 2022-04-04
Use `bigint` within `sql` template string query


nicholaschiang:
## Desired Behavior

I want to be able to include `bigint`s in my `sql` template string queries like this:

![image](https://user-images.githubusercontent.com/20798889/161466731-ecf7ab83-b857-42f4-993d-78085cbf57d4.png)

## Motivation

Right now, `bigint`s are a pain to manage when interacting with Postgres from a Node.js environment. This is why I switched from using Prisma's `db.$queryRaw` method (see prisma/prisma#8121 and prisma/prisma#12367) which doesn't support `bigint` serialization and de-serialization very well. It appeared from [this library's documentation](https://github.com/gajus/slonik#built-in-type-parsers) that `bigint` de-serialization is already handled, and thus I assumed that serialization (i.e. using a `bigint` in a `sql` template string query) would also be handled. But it wasn't (at least AFAIK). Which is why I'll probably be [switching to `pg-promise` for now](https://github.com/vitaly-t/pg-promise/issues/657).

## Implementation

Copy over the `pg-promise` implementation (see vitaly-t/pg-promise#657 that's been around since 2019).

gajus:
Related to https://github.com/gajus/slonik/pull/299

gajus:
The problem is that if we start allowing BigInt, as an input and as an output, then you need to use them consistently across your application anywhere where the value can be a BigInt, which is.. most of your code base likely.

I don't see a good design for the addition of bigint for that reason, but I am open to proposals.

nicholaschiang:
@gajus that's exactly what I'm doing though. I use Prisma and define my ID columns there as `BigInts` and then use those Prisma Typescript types across my code base, ensuring that I'm treating every ID as a `BigInt` everywhere. 

The problem arises because Prisma's support for `BigInts` in raw queries is lacking (see prisma/prisma#12367) and thus I need to use another Node.js Postgres client library while they're still working on polishing that up. I came first to `slonik` because it seemed like an interesting alternative (that, like Prisma's `$queryRaw` features, also uses ES6 template strings), but was disappointed by this lack of support for `BigInts`.

I think it makes sense to represent a `BigInt` from your database as a `BigInt` in your application code (as long as the language your application code is written in supports it... and as of Node.js 10, it does).

gajus:
Let's see how this survey goes https://twitter.com/kuizinas/status/1561734172938469378

AndrewJo:
[...]

Issue #184, opened 2020-06-11
Should sql.array accept `bigint` as the value of `memberType`?


gavacho:
The `memberType` argument of `sql.array` is typed as `TypeNameIdentifierType`:

```ts
export type TypeNameIdentifierType =
  'bool' |
  'bytea' |
  'float4' |
  'float8' |
  'int2' |
  'int4' |
  'json' |
  'text' |
  'timestamptz';
```

Which excludes `bigint`. Is that intentional? It seems like a potential oversight.

gajus:
Does node-postgres even support bigint?

gajus:
Just in case, this would be `int8`, not `bigint`. See [comments about keyword vs type name](https://github.com/gajus/slonik#sqlarray-membertype).

gavacho:
Yeah, we're using bigint columns. Aren't numbers in js 8 bytes?

hjr3:
The ask here is for `TypeNameIdentifierType` to accept `int8`. An 8byte integer is compatible with JS `Number` type. This is not asking for [BigInt in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) support.

nonara:
> The ask here is for `TypeNameIdentifierType` to accept `int8`. An 8byte integer is compatible with JS `Number` type. This is not asking for [BigInt in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) support.

Correct. I can confirm using `<any>'int8'` works. This can be closed by adding `int8` to the `TypeNameIdentifierType` union

PR #510, opened 2023-09-28
feat: add bigint support (fixes #454 #338 #184)

### LINKED ISSUE #454: Using `bigint` in `sql.array` results in TypeScript compiler error
When using a table with schema that contains a **BIGINT**/**INT8** column, trying to use `sql.array` on an array of `bigint` primitives (e.g. `sql.array([9007199254740999n, 9007199254741000n, 9007199254741001n], "bigint")`) results in `Type 'bigint' is not assignable to type 'PrimitiveValueExpression'.` error.

## Expected Behavior

TypeScript should compile the code.

## Current Behavior

TypeScript compiler error:

```
Type 'bigint' is not assignable to type 'PrimitiveValueExpression'.
```

## Possible Solution

Unless we plan on supporting EOL Node.js versions (probably should be a separate discussion on whether the library should support only active LTS versions), `bigint` has been a stable feature in Node.js for years at this point. Maybe it should be one of the union types for `PrimitiveValueExpression`.

If there aren't any concerns with unioning `bigint` with rest of the primitive types in `PrimitiveValueExpression`, I'll be happy to submit a PR to fix this issue.

## Steps to Reproduce

1. ```typescript
   const ids = [9007199254740999n, 9007199254741000n, 9007199254741001n];
   const sqlArrayFragment = sql.array(ids, "int8");
   ```
2. ```
   error TS2345: Type 'bigint[]' is not assignable to type 'readonly PrimitiveValueExpression[]'.
     Type 'bigint' is not assignable to type 'PrimitiveValueExpression'.

   2 const sqlArrayFragment = sql.array(ids, "int8");
                                        ~~~

   Found 1 error.
   ```

gajus:


coreyward:
Not sure if this is anticipated behavior, but it seems that this affects aggregates of non-int8 columns (e.g. `sum` or `count`). I think this makes some sense technically, but it’s a challenging change to make confidently across a large codebase.

PR #522, opened 2023-10-25
Test unexpected termination

gajus:


gajus:
Working on properly handling unexpected backend termination

https://github.com/brianc/node-postgres/issues/3083

PR #530, opened 2024-01-18
Fix bigint in sql.join

### LINKED ISSUE #529: Using values with `bigint`s in `sql.join` throws `Unexpected value expression.`
An example being worth a thousand words:

```typescript
const a = sql.unsafe`a = ${1n}`; // ok
const b = sql.unsafe`b = ${2n}`; // ok
const joined = sql.join([a, b], sql.unsafe` AND `); // ok
const result = sql.unsafe`${joined}`; // error !
```

image

related issue: #338

oguimbal:
Fixes #529

### DIFF
--- a/src/factories/createPrimitiveValueExpressions.ts
+++ b/src/factories/createPrimitiveValueExpressions.ts
@@ -12,7 +12,7 @@ export const createPrimitiveValueExpressions = (
 ): readonly PrimitiveValueExpression[] => {
   const primitiveValueExpressions: Array<
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    any[] | Buffer | boolean | number | string | null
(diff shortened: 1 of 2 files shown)


oguimbal:
linting issue should be fixed

oguimbal:
FYI I deployed [the fixed version on npm](https://www.npmjs.com/package/@oguimbal/slonik) if someone needs it (we did)... temparily replace your current slonik package by `"slonik": "npm:@oguimbal/slonik",`  in package.json and you'll be good to go after a `yarn` / `bun i` / `node i` etc

gajus:
I was hesitant to merge this because I remember going down this path and it backfired, but I cannot remember exactly how. Merging this until someone raises an issue about it.

PR #534, opened 2024-02-15
feat: Add field names in mockQueryResult

### LINKED ISSUE #232: createMockQueryResult and transformRow incompatibility
I'm not sure whether this is the right place for this issue or if it would be better placed in [slonik-interceptor-field-name-transformation](https://github.com/gajus/slonik-interceptor-field-name-transformation). I'm happy to move it if you'd prefer.

`createMockQueryResult` only sets the `rows` and `rowCount` fields of the mocked result and stubs the rest, including `fields`. However, `transformRow` interceptors receive the `fields` as a parameter and `slonik-interceptor-field-name-transformation` at least is incompatible with `createMockQueryResult` due to that.

## Expected Behavior

`slonik-interceptor-field-name-transformation` should transform the mocked result set as usual.

## Current Behavior

`slonik-interceptor-field-name-transformation` returns empty objects because it was not told about any fields.

## Possible Solution

I see two or three possible solutions, really:

1. Mock `fields` within `createMockQueryResult` so that `transformRow` can behave as usual; this would allow any other interceptors using that hook to function as expected.
1. Change `slonik-interceptor-field-name-transformation` to not rely on `fields` for transformation, and instead transform the keys of the objects in `rows` directly.
1. Allow callers to pass in `fields` as well, in some form. This feels a bit hacky and is my least preferred option, for sure.

I am currently working around the issue by mocking `fields` myself before returning the mocked result (i.e. first option above).

## Steps to Reproduce

1. Use `slonik-interceptor-field-name-transformation` in `createMockPool()`
1. Make a query
1. Observe result set with empty object(s)

Reproduction adapted from the example in the README:

```javascript
import {
  createMockPool,
  createMockQueryResult,
} from 'slonik';
import { createFieldNameTransformationInterceptor } from 'slonik-interceptor-field-name-transformation'

const pool = createMockPool(
  {
    query: async () => {
      return createMockQueryResult([
        {
          foo: 'bar',
        },
      ]);
    },
  },
  {
    interceptors: [
      createFieldNameTransformationInterceptor({ format: 'CAMEL_CASE' }),
    ],
  }
);

await pool.connect(async (connection) => {
  const results = await connection.query(sql`
    SELECT ${'foo'}
  `);
  console.log(results);

  // {
  //  command: 'SELECT',
  //  fields: [],
  //  notices: [],
  //  rowCount: 1,
  //  rows: [ {} ]
  // }
});
```

## Logs

Probably not relevant here, but I guess I can get some if you'd like 😀

ovstetun:
Take all keys from first row, unspecified datatype (0). (fixes #232)

### DIFF
--- a/src/factories/createMockQueryResult.ts
+++ b/src/factories/createMockQueryResult.ts
@@ -1,11 +1,16 @@
-import { type QueryResult, type QueryResultRow } from '../types';
+import { type Field, type QueryResult, type QueryResultRow } from '../types';
 
 export const createMockQueryResult = (
   rows: readonly QueryResultRow[],
 ): QueryResult<QueryResultRow> => {
+  let fields: Field[] = [];
+  if (rows.length > 0) {
+    fields = Object.keys(rows[0]).map((it) => ({ dataTypeId: 0, name: it }));
+  }
+
   return {
     command: 'SELECT',
-    fields: [],
+    fields,
     notices: [],
     rowCount: rows.length,
     rows,


gajus:
Sorry for not seeing this earlier. Makes sense.

Issue #539, opened 2024-03-07
Improving Runtime Validation Performance through Async Parsing


janpaepke:
Hi @gajus,
as [you noticed yourself](https://github.com/colinhacks/zod/issues/205#issuecomment-1285659801), type validation using zod is fairly CPU hungry.  
I also noticed this, particularly with nested shapes and combined with transforms.

Without optimizing zod, I don't think there's much we can do about this.

There is however another optimisation that can be done with little effort inside of slonik:
**Allowing row transformations to be async**.
This may not decrease CPU load, but will likely make services more responsive, as outlined below.

## Motivation

Recently I noticed large spikes in the Eventloop lag in our production environment (upwards of 4 seconds).
This is obviously bad, since it blocks any other requests to the service from being processed.

I was able to track the origin down to a sync operation which had both: A complex result shape (nested jsonb) as well as a relatively large dataset. 
At 0.2 -  1ms parsing time per row, the numbers quickly grew.

As a solution I used `.stream()`, rather than `many()`.

This had two effects: The operation was spread over a longer period (which was fine), but more importantly the parsing was done asynchronously, in batches, which meant it didn't freeze up the service.

## Implementation

After looking into this some more, I noticed that all row transformations are processed synchronously in `executeQuery`.  
From what I can tell, there is really no reason for this, as `executeQuery` is obviously async already.

So here's my Proposal:
1. Change the result type of `transformRow` to `MaybePromise<QueryResultRow>`.
2. in `executeQuery` map over rows and use `Promise.all` **(see below)**
3. in `stream` update the `transform` function passed into `Transform` to do the same and make it async: `async transform(datum, enc, callback)`. As far as I can tell this will not change the behaviour and will call `callback();` once all rows have been asynchronously transformed.
4. add tests for promise results of `transformRow`.
5. update docs to recommend `z.safeParseAsync` over `z.safeParse`: `const validationResult = await resultParser.safeParseAsync(row);`

This is how I'd process the transformations in `executeQuery`:
```ts
for (const interceptor of interceptors) {
  if (interceptor.transformRow) {
    const { transformRow } = interceptor;
    const { fields } = result;

    const rows: QueryResultRow[] = await Promise.all(
      result.rows.map(
        row => transformRow(executionContext, actualQuery, row, fields)
      )
    );

    result = {
      ...result,
      rows,
    };
  }
}
```

This change would be fully backwards compatible.

[...]

PR #540, opened 2024-03-08
Make row transforms async

janpaepke:
This implements the changes outlined in #539.

I also took the liberty of pulling up the "Result parser interceptor" section to the beginning of the runtime_validation section.
This felt reasonable, since this is the initial setup without which none of the following info would work.

I realise this makes reviewing the changes a bit more difficult, so here they are:
- [line 24](https://github.com/gajus/slonik/pull/540/files#diff-0f9cfa6ab7e48e7434ebbb29e5561db27af0193ce3830c703eb05360b6e826beR24): added "based on" as it's not exactly the original implementation anymore.
- [line 39](https://github.com/gajus/slonik/pull/540/files#diff-0f9cfa6ab7e48e7434ebbb29e5561db27af0193ce3830c703eb05360b6e826beR39): added 'async'
- [line 46](https://github.com/gajus/slonik/pull/540/files#diff-0f9cfa6ab7e48e7434ebbb29e5561db27af0193ce3830c703eb05360b6e826beR46): added comment
- [line 47](https://github.com/gajus/slonik/pull/540/files#diff-0f9cfa6ab7e48e7434ebbb29e5561db27af0193ce3830c703eb05360b6e826beR47): changed sync to async implementation

When looking more closely at the readme, I also noticed that the "Unknown keys" section seems to be outdated.
Zod does not fail on unknown keys by default, it just strips them.
I suspect this section is a relict from the brief time when zod was hard-wired into slonik.

I will create a [separate PR](https://github.com/gajus/slonik/pull/541) for this.

### DIFF
--- a/.README/INTERCEPTORS.md
+++ b/.README/INTERCEPTORS.md
@@ -54,7 +54,7 @@ type Interceptor = {
   beforeTransformQuery?: (
     queryContext: QueryContext,
     query: Query
-  ) => Promise<null>,
+  ) => MaybePromise<null>,
   queryExecutionError?: (
     queryContext: QueryContext,
     query: Query,
@@ -69,7 +69,7 @@ type Interceptor = {
     query: Query,
     row: QueryResultRow,
     fields: Field[],
-  ) => QueryResultRow
+  ) => MaybePromise<QueryResultRow>
 };
 ```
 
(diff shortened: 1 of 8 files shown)


gajus:
Thank you

Issue #548, opened 2024-03-22
Cannot use $1 in nested TEXT


alxndrsn:
## Versions

* `slonik`: `37.4.0`
* `node`: `18.18.2`

## Observed

Including dollar sign + a number in a nested `sql.fragment()` throws an `Error`.

## Expected

Should be fine.

## Repro

```js
const { sql } = require('slonik');

const query0 = sql.fragment`UPDATE item SET price='$5' WHERE id=1`;
const query1 = sql.fragment`${query0} RETURNING *`;
```

### Logs

```
./node_modules/slonik/dist/sqlFragmentFactories/createFragmentSqlFragment.js:20
        throw new errors_1.UnexpectedStateError('The greatest parameter position is greater than the number of parameter values.');
        ^
UnexpectedStateError: The greatest parameter position is greater than the number of parameter values.
    at createFragmentSqlFragment (./node_modules/slonik/dist/sqlFragmentFactories/createFragmentSqlFragment.js:20:15)
    at createSqlTokenSqlFragment (./node_modules/slonik/dist/factories/createSqlTokenSqlFragment.js:28:74)
    at createFragment (./node_modules/slonik/dist/factories/createSqlTag.js:39:91)
    at Object.fragment (./node_modules/slonik/dist/factories/createSqlTag.js:83:20)
    at Object.<anonymous> (./index.js:4:28)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:86:12) {
  cause: undefined
}
```

gajus:
Is the intent to record a literal `$1` value? If yes, you will need to do it as:

```ts
sql.fragment`UPDATE item SET price=${'$5'} WHERE id=${1}`;
```

This is not a bug, but a design choice.

PR #549, opened 2024-03-22
bug: cannot handle $\d in nested fragments

alxndrsn:
Test cases for https://github.com/gajus/slonik/issues/548


gajus:
Provided context in https://github.com/gajus/slonik/issues/548.

PR #552, opened 2024-03-23
bug: cannot handle $\d anywhere in SQL statements

alxndrsn:


gajus:
@alxndrsn Tests are passing. Just need to fix linting issue and this is good to merge.

PR #554, opened 2024-03-25
feat: use custom placeholders (fixes #553)

### LINKED ISSUE #553: Cannot use $1 or $\d+ in nested SQL fragments
More general version of https://github.com/gajus/slonik/issues/548.

Test cases at https://github.com/gajus/slonik/pull/552.

I appreciate per https://github.com/gajus/slonik/issues/548 that this is a design decision, so feel free to close immediately.  I think it's helpful to have here as documentation in case others run into this issue in future.  Also I may be mis-using the API?

## Observed

Cannot use $-number combinations (e.g. `$1`) anywhere in nested SQL fragments.

## Expected

Valid postgres-flavoured SQL would be fine; template strings wouldn't be parsing the string parts of a template.

## Repro

* https://github.com/gajus/slonik/pull/552

gajus:


gajus:
CC @alxndrsn FYI, experimenting with a solution for https://github.com/gajus/slonik/issues/553

gajus:
Pleasantly surprised that this passes all the tests.

PR #559, opened 2024-03-29
Evaluate monorepo setup

gajus:
The primary goal is to separate drivers (like `pg`) from Slonik package. This would allow a better support for alternative drivers, e.g. https://www.npmjs.com/package/postgres.

This change would also mean that minimum Slonik setup would look as follows:

```ts
import {
  createPgDriverFactory,
} from '@slonik/pg-driver';
import {
  createPool,
} from 'slonik';

// Slonik itself does not provide a driver implementation.
// To use Slonik with PostgreSQL, you must use a driver implementation,
// e.g. the `pg` driver is available at https://www.npmjs.com/package/@slonik/pg-driver.
const driverFactory = createPgDriverFactory();

const pool = await createPool('postgres://', {
  driverFactory,
});
```

The other nice thing about this setup is that it makes it very easy to develop compatible drivers.

```ts
import { createPgDriverFactory } from '../factories/createPgDriverFactory';
import { createIntegrationTests, createTestRunner } from '@slonik/internals';

const driverFactory = createPgDriverFactory();

const { test } = createTestRunner(driverFactory, 'pg');

createIntegrationTests(test, driverFactory);
```

`createIntegrationTests` can now be easily used to ensure compatibility of a custom driver with Slonik.

This is not necessary an unlock of a monorepo setup, but it makes the separation more obvious.

TODO:

* [ ] automate releases using changesets


gajus:
I have not found a good way to do this separation. Quickly ended up with dozens of internal packages with little value.

PR #566, opened 2024-04-15
fix: wait for release routine to run the clean up query

gajus:


gajus:
Would be nice to add a test for this scenario. I was not able to write one, but I was able to catch this scenario in our codebase.

PR #595, opened 2024-05-08
tokens: use Symbols instead of strings

alxndrsn:


gajus:
I am not against, but can you provide reasoning for why/how this is better?

gajus:
Separately, this highlights issue with the current test suite, as this would definitely break the [slonik-sql-tag-raw](https://github.com/gajus/slonik/tree/main/packages/slonik-sql-tag-raw) package.

mmkal:
This will likely also cause problems for anyone trying to bundle slonik with esbuild, swc, webpack, turbopack, bun, parcel, ncc, etc. etc.

gajus:
> This will likely also cause problems for anyone trying to bundle slonik with esbuild, swc, webpack, turbopack, bun, parcel, ncc, etc. etc.

aligned. I've seen this also cause issues when multiple versions of Slonik are loaded.

I can see why this could be desirable, but also cautious about how it would need to be implemented.

alxndrsn:
> this would definitely break the [slonik-sql-tag-raw](https://github.com/gajus/slonik/tree/main/packages/slonik-sql-tag-raw) package.

> This will likely also cause problems for anyone trying to bundle slonik...

> I've seen this also cause issues when multiple versions of Slonik are loaded.

@gajus @mmkal can you expand on likely issues?  I can see how all of these things could be problematic if using the [`Symbol()` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/Symbol), but not with [`Symbol.for()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for).

gajus:
> @gajus @mmkal can you expand on likely issues? I can see how all of these things could be problematic if using the [Symbol() constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/Symbol), but not with [Symbol.for()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for).

My oversight. I was not familiar with `Symbol.for()`.

This has now been released https://github.com/gajus/slonik/releases/tag/%40slonik%2Futilities%4044.0.0

PR #620, opened 2024-06-06
Parse queries in Slonik

gajus:


packages/slonik/src/factories/createQueryValidator.ts
```diff
@@ -0,0 +1,41 @@
...
+import { parseAsync } from 'pgsql-parser';
+
+export const createQueryValidator = () => {
+  const queries: Record<string, number> = {};
+
+  return async (query: Query) => {
+    let stats = queries[query.sql];
+
+    if (stats === undefined) {
+      const ast = await parseAsync(stripSlonikPlaceholders(query.sql));
```
danielrearden, line 13: @gajus you're calling this twice -- once here and again inside `executeQuery` before this is called

danielrearden, line 13: We may want to consider what happens if `parseAsync` fails.

Minimally, if the query's invalid, we should make sure the error that's thrown includes all the details needed to determine the syntax issue.

However, you might also consider letting invalid queries through to the server -- one, that would ensure the syntax errors are the same as what we get today and two, that would avoid issues where the parser is not up to date and we try to parse some newer syntax.

gajus: Added test to cover this scenario. Thank you

gajus: Refactored

PR #636, opened 2024-08-07
Fix readme ToC links

mk0x9:
Many links in README's ToC to slonik's methods are broken. As an extra, I added dots to the features list on missing points.


gajus:
Thank you

PR #641, opened 2024-08-26
Align sslMode with current libpq-connect options

wasd171:
This PR aligns TypeScript types with [current](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-SSLMODE) `libpq-connect` options

### DIFF
--- a/.changeset/hot-shoes-destroy.md
+++ b/.changeset/hot-shoes-destroy.md
@@ -1,0 +1,6 @@
+---
+"@slonik/types": major
+---
+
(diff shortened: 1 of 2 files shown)


wasd171:
@gajus I would still like to move forward with this PR, but `no-verify` seems like more of a product decision. Would you prefer to retain it or can I remove it not only from the types, but also from the codebase?

gajus:
Happy to remove it.

Will ship as a major version upgrade since it is breaking change.

PR #661, opened 2024-12-20
fix: pending connection handling

mikeroelens:
Was looking through connection pool code and noticed a few things that _may_ cause problems. I haven't proved them out in practice on impact but seems like without these fixes the pool could technically end up in a bad state

https://www.loom.com/share/563a05d796404378ac418580ba65038e


packages/slonik/src/factories/createConnectionPool.ts
```diff
@@ -144,22 +144,25 @@ export const createConnectionPool = ({
       pendingConnections.push(pendingConnection);
 
       const connection = await pendingConnection.catch((error) => {
-        pendingConnections.pop();
```
mikeroelens, line 147: `pendingConnections.pop();` will remove the most recently added pending connection from the array. But by the time this code runs the last connection in the array might not be the same one that threw an error right?

E.g.
- Pending Connection A added. `pendingConnections = [A]`
- Pending Connection B added. `pendingConnections = [A,B]`
- 5s later A has timed out and throws an error. `pendingConnections.pop()` is called and now B is removed and we’re left with `[A]`. But it should've been `A` that was removed

packages/slonik/src/factories/createConnectionPool.ts
```diff
@@ -144,22 +144,25 @@ export const createConnectionPool = ({
...
+        pendingConnections.splice(
+          pendingConnections.indexOf(pendingConnection),
+          1,
+        );
 
         throw error;
       });
 
       const onRelease = () => {
+        if (connection.state() !== 'IDLE') {
```
mikeroelens, line 156: I moved this up further. If not technically we might call `waitingClients.shift()` but then never actually acquire a connection for that waiting client since we throw an error.

packages/slonik/src/factories/createConnectionPool.ts
```diff
@@ -173,14 +176,13 @@ export const createConnectionPool = ({
 
         connections.splice(connections.indexOf(connection), 1);
 
-        const waitingClient = waitingClients.shift();
```
mikeroelens, line 176: Same thing here. We’re calling waitingClients.shift() but then if for some reason we’re below the minimum pool size we return early and the waiting client never gets a connection.

I removed the `return` after `addConnection()`, I don't see an obvious reason it means we need to return early but could be missing something

packages/slonik/src/factories/createConnectionPool.ts
mikeroelens: This is causing some tests to fail because it's throwing `Connection is not idle.`. From logging it looks like the state is actually `PENDING_DESTROY`. Which I think indicates that this is a true problem in practice, we shouldn't be calling `waitingClients.shift()` if we can't actually do something with that waiting client.

Maybe we should change this to:
```
const onRelease = () => {
        if (connection.state() !== 'IDLE') {
            return;
        }

        ...
```
?

gajus:
@mikeroelens any chance of getting this across the line?

packages/slonik/src/factories/createConnectionPool.ts
mikeroelens: Was able to reproduce this behaviour with a test. Although it also required making [this change](https://github.com/gajus/slonik/pull/661/commits/aea780c8d0d0d0aa13947f6353dab51e3e9d5a00) so that we don't accidentally call `splice` with `-1`. This was somewhat hiding the failure because it was still removing an item from `pendingConnections` even though it wasn't the correct one.

packages/slonik/src/types.ts
```diff
@@ -360,6 +360,7 @@ type ConnectionContext = {
 type PoolState = {
   readonly acquiredConnections: number;
   readonly idleConnections: number;
+  readonly pendingConnections: number;
```
mikeroelens, line 363: I added this so we can see how many connections are pending. This was required for an integration test case but also seems like a good thing to have anyways

gajus:
This is amazing!

PR #666, opened 2025-01-03
fix: add missing ssl models

gajus:


SylvainMarty:
Hey @gajus, do you plan to release these changes anytime soon ?

The maintainer of the `pg` lib [adviced me not so long ago](https://github.com/brianc/node-postgres/issues/3420#issuecomment-2800186308) that we should use `sslmode=verify-full` instead of `sslmode=require` as it has been deprecated.

On the latest version of Slonik, providing `sslmode=verify-full` throws a Zod validation error and I guessed this PR is trying to fix that. 😄

gajus:
Happy to release it, just don't have confidence it does what's desired. Take a minute to review the code if you are familiar with the domain, and if yes, I will release it.

packages/utilities/src/utilities/parseDsn.ts
```diff
@@ -56,7 +57,17 @@ export const parseDsn = (dsn: string): ConnectionOptions => {
         .describe(
           'Specifies the location for the secret key used for the client certificate.',
         ),
-      sslmode: z.enum(['disable', 'no-verify', 'require']).optional(),
+      sslmode: z
+        .enum([
+          'allow',
```
SylvainMarty, line 62: `allow` is not listed in [`pg` documentation](https://github.com/brianc/node-postgres/blob/master/packages/pg-connection-string/README.md#tcp-connections) so I don't think we should support it.

Also, it is not used in the switch line 147.

packages/types/src/types.ts
```diff
@@ -8,13 +8,22 @@ export type ConnectionOptions = {
...
+  ssl?:
+    | boolean
+    | {
+        ca?: string;
+        cert?: string;
+        key?: string;
+        rejectUnauthorized: boolean;
+      };
+  sslMode?:
+    | 'allow'
```
SylvainMarty, line 20: Same comment here

SylvainMarty (commented):
From the PR https://github.com/brianc/node-postgres/pull/2709 mentioned in the code (which has been merged 3 days ago), they [introduced a new flag](https://github.com/brianc/node-postgres/blob/master/packages/pg-connection-string/README.md#tcp-connections)  to actually use the SSL options the same way as libpq.

More importantly, they added this warning at the bottom of the doc:
> [!CAUTION]
> Choosing an sslmode other than verify-full has serious security implications. Please read https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS to understand the trade-offs.

Reading this, I think Slonik should use the new `sslcompat=libpq` flag and implement the new SSL options (that will be the one used by default in `pg` one day, see https://github.com/brianc/node-postgres/issues/3421). In the doc we should also advocate the same way about the trade-offs.

Since Slonik only supported 3 limited SSL options before, IMO it is an acceptable change. Sadly `sslmode=no-verify` might have to be dropped since it is not supported by `libpq` (or adapted to an other flag internally).

SylvainMarty:
Since Slonik is built on top of `pg`, why don't we directly pass the connection string to `pg`?

I have the impression the driver system of Slonik was a way to not close any doors so one day we can implement a connector that uses a different library than `pg` (or maybe even to another RDBMS).
But since there is a whole `createPgDriverFactory()` function we can already use, why don't we pass a connection string or a config object directly to the driver factory parameters instead of passing them to the `createPool()` function?
I mean, each driver might have their specific options so it would make sense the driver factory parameters is a facade to the internal library parameters, no?

That would mean removing some complex security logic from Slonik and trusting the underlying library.
To me this is not a bad idea: we can clearly see some people are spending a huge amount of time and energy on `pg` so it follows better security practices.

What do you think?

PR #724, opened 2025-10-29
docs: enhance field-name interceptor package documentation

gabrielrra:
This change makes the documentation more clear on how to install the library and what it does


gajus:
Thank you

PR #741, opened 2025-12-12
Use "unknown" instead of "any"

### LINKED ISSUE #740: sql.unsafe produces results of type "any / any[]" that triggers @typescript-eslint/no-unsafe-assignment
This is the current definition of the `QuerySqlToken`

https://github.com/gajus/slonik/blob/a9a02c0e05c0af5c98e7afb5ad1a50df14637801/packages/sql-tag/src/types.ts#L65-L73

And the current definition of the `sql.unsafe`:

https://github.com/gajus/slonik/blob/a9a02c0e05c0af5c98e7afb5ad1a50df14637801/packages/sql-tag/src/factories/createSqlTag.ts#L234-L240

As a result, queries that are constructed via the `sql.unsafe` tags return results of `any` / `any[]`, etc. A few releases ago we were getting `readonly any[]` types which did not trigger the `@typescript-eslint/no-unsafe-assignment` rule

## Desired Behavior

I would prefer if slonik could return results of `unknown` / `unknown[]` types for such queries. This would still force the developer to do the runtime type validation (we use [typebox](https://github.com/sinclairzx81/typebox) instead of zod) but with the additional type-checking [upsides](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type)

## Implementation

It would probably be enough to do the following:

```ts
// sql-tag/src/types.ts
T extends StandardSchemaV1 = StandardSchemaV1<unknown, unknown> // instead of T extends StandardSchemaV1 = StandardSchemaV1<any, any>
```

```ts
// sql-tag/src/factories/createSqlTag.ts
parser: z.unknown()  // instead of parser: z.any()
```

wasd171:
This should fix https://github.com/gajus/slonik/issues/740

### DIFF
--- a/.changeset/public-kings-like.md
+++ b/.changeset/public-kings-like.md
@@ -1,0 +1,6 @@
+---
+"@slonik/sql-tag": major
+"slonik": major
+---
+
+Replaced `any` output types (e.g. for `sql.unknown`) with `unknown`
--- a/README.md
+++ b/README.md
@@ -2096,7 +2096,7 @@ Produces:
 (template: TemplateStringsArray, ...values: ValueExpression[]) => QuerySqlToken;
 ```
 
-Creates a query with Zod `any` type. The result of such a query has TypeScript type `any`.
+Creates a query with Zod `unknown` type. The result of such a query has TypeScript type `unknown`.
 
 ```ts
 const result = await connection.one(sql.unsafe`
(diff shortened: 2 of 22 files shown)


wasd171:
^ cannot run pnpm commands because `slonik@48.8.8` is not published to NPM

gajus:
Sorry for overlooking this. Let me know if you're still planning on updating.

wasd171:
@gajus could you please trigger the pipeline?

wasd171:
@gajus pushed an update, hopefully this would work with the latest changes from `main`

wasd171:
@gajus I've rebased this PR, ready for review

wasd171:
@gajus do you have capacity to check this PR? would be happy to adapt it, if needed

gajus:
This is getting released in https://github.com/gajus/slonik/pull/791.

